PR #1, opened 2022-02-17
DPE-48 Adding basic charm library

WRFitch:
This PR creates a charm library that abstracts some common functions across the pgbouncer charms, which will be published to the /edge channel of the pgbouncer-operator charm on charmhub once implemented. 

The functions this library abstracts are as follows:
- Generating passwords
- Generating config files 

These will obviously expand over time as the charms become more fully featured.

[Jira Ticket](https://warthogs.atlassian.net/browse/DPE-48?atlOrigin=eyJpIjoiZjY4ZmNmZWI5YjhhNDNjMWIzMjE5Mjg2YWNiZDA0YzAiLCJwIjoiaiJ9)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,5 +1,9 @@
 name: Tests
-on: [pull_request, push]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
   lint:
@@ -33,12 +37,13 @@ jobs:
         with:
           provider: microk8s
       - name: Run integration tests
-        # set a predictable model name so it can be consumed by charm-logdump-action
-        run: tox -e integration -- --model itest
-      - name: Dump logs
-        uses: canonical/charm-logdump-action@main
-        if: failure()
-        with:
-          # TEMPLATE-TODO: Replace the application name
-          app: operator-template
-          model: itest
+        # TODO set a predictable model name so it can be consumed by charm-logdump-action
+        # For now, setting one using "-- --model testing" causes it to get lost in the CI. This can be replicated locally.
+        run: tox -e integration
+      # - name: Dump logs
+      #   uses: canonical/charm-logdump-action@main
+      #   if: failure()
+      #   with:
+      #     # TEMPLATE-TODO: Replace the application name
+      #     app: operator-template
+      #     model: testing
--- a/lib/charms/dp_pgbouncer_operator/v0/pgb.py
+++ b/lib/charms/dp_pgbouncer_operator/v0/pgb.py
@@ -23,9 +23,26 @@ from typing import Dict
 
 from ops.model import ConfigData
 
+PGB_INI = """\
+[databases]
+{databases}
+[pgbouncer]
+listen_port = {listen_port}
+listen_addr = {listen_addr}
+auth_type = md5
+auth_file = userlist.txt
+logfile = pgbouncer.log
+pidfile = pgbouncer.pid
+admin_users = {admin_users}
+"""
+
 
 def generate_password() -> str:
-    """Generates a secure password.
+    """Generates a secure password of alphanumeric characters.
+
+    Passwords are alphanumeric only, to ensure compatibility with the userlist.txt format -
+    specifically, spaces and double quotes may interfere with parsing this file.
+
     Returns:
         A random 24-character string of letters and numbers.
     """
@@ -44,16 +61,12 @@ def generate_pgbouncer_ini(users: Dict[str, str], config: ConfigData) -> str:
     Returns:
         A multiline string defining a valid pgbouncer.ini file
     """
-    return f"""[databases]
-{config["pgb_databases"]}
-[pgbouncer]
-listen_port = {config["pgb_listen_port"]}
-listen_addr = {config["pgb_listen_address"]}
-auth_type = md5
-auth_file = userlist.txt
-logfile = pgbouncer.log
-pidfile = pgbouncer.pid
-admin_users = {",".join(users.keys())}"""
+    return PGB_INI.format(
+        databases=config["pgb_databases"],
+        listen_port=config["pgb_listen_port"],
+        listen_addr=config["pgb_listen_address"],
+        admin_users=",".join(users.keys()),
+    )
 
 
 def generate_userlist(users: Dict[str, str]) -> str:
@@ -66,3 +79,27 @@ def generate_userlist(users: Dict[str, str]) -> str:
         space, one pair per line.
     """
     return "\n".join([f'"{username}" "{password}"' for username, password in users.items()])
+
+
+def parse_userlist(userlist: str) -> Dict[str, str]:
+    """Parse userlist.txt into a dictionary of usernames and passwords.
+
+    Args:
+        userlist: a multiline string of users and passwords, formatted thusly:
+        '''
+        "test-user" "password"
+        "juju-admin" "asdf1234"
+        '''
+    Returns:
+        users: a dictionary of usernames and passwords
+    """
+
+    parsed_userlist = {}
+    for line in userlist.split("\n"):
+        if line.strip() == "" or len(line.split(" ")) != 2:
+            continue
+        # Userlist is formatted "{username}" "{password}""
+        username, password = line.replace('"', "").split(" ")
+        parsed_userlist[username] = password
+
+    return parsed_userlist
--- a/lib/charms/dp_pgbouncer_operator/v0/pgb.py
+++ b/lib/charms/dp_pgbouncer_operator/v0/pgb.py
@@ -1,0 +1,62 @@
+# Copyright 2022 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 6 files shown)


lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -0,0 +1,68 @@
...
+"""
+
+import secrets
+import string
+from typing import Dict
+
+from ops.model import ConfigData
+
+
+def generate_password() -> str:
```
quiet-ranger, line 27: This function looks generic enough to me and useful to others.  We should consider extracting it to a generic library with security related functions.

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -0,0 +1,68 @@
...
+
+    This is a basic stub method, and will be updated in future to generate more complex
+    pgbouncer.ini files in a more sophisticated way.
+
+    Args:
+        users: a dictionary of usernames and passwords
+    Returns:
+        A multiline string defining a valid pgbouncer.ini file
+    """
+    return f"""[databases]
```
quiet-ranger, line 47: Do we plan to re-implement this with, say, templates?  It is understandable but somewhat difficult to read, especially if this is expected to grow with many more configuration items.

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -0,0 +1,68 @@
...
+def generate_userlist(users: Dict[str, str]) -> str:
+    """Generate userlist.txt from the given dictionary of usernames:passwords.
+
+    Args:
+        users: a dictionary of usernames and passwords
+    Returns:
+        A multiline string, containing each pair of usernames and passwords separated by a
+        space, one pair per line.
+    """
+    return "\n".join([f'"{username}" "{password}"' for username, password in users.items()])
```
quiet-ranger, line 68: This is a one liner using dictionary comprehension and a bit of string formatting.  Does it really need to be a separate function?

tests/unit/test_pgb.py
```diff
@@ -0,0 +1,42 @@
...
+
+import unittest
+
+from lib.charms.dp_pgbouncer_operator.v0 import pgb
+
+
+class TestPgb(unittest.TestCase):
+    def test_generate_password(self):
+        pw = pgb.generate_password()
+        self.assertEqual(len(pw), 24)
```
quiet-ranger, line 12: Very good.  But can we also ensure that the resulting string contains only characters and digits?  Also, it would be good to write a comment in the implementation itself explaining why we made this choice and, for example, did not include some punctuation characters.

tests/unit/test_pgb.py
```diff
@@ -0,0 +1,42 @@
...
+        # Though this isn't correctly mocking an ops.model.ConfigData object, ConfigData implements
+        # a LazyMapping under the hood that accesses variables in the same way as a dictionary -
+        # they're effectively interchangeable in this context.
+        config = {
+            "pgb_databases": "test-dbs",
+            "pgb_listen_port": "4454",
+            "pgb_listen_address": "4.4.5.4",
+        }
+        pgb_ini = pgb.generate_pgbouncer_ini(users, config)
+        expected_pgb_ini = f"""[databases]
```
quiet-ranger, line 25: This implementation is fragile.  If someone changes the order or adds a new element in the implementation, the unit test will fail.  Please rethink this one.

quiet-ranger (changes requested):

lib/charms/dp_pgbouncer_operator/v0/pgb.py
WRFitch: This sounds useful. I'm also planning to abstract the encryption and hashing logic for pgbouncer authentication into this library, so perhaps this library can be abstracted once it's more than a single small function ;)

lib/charms/dp_pgbouncer_operator/v0/pgb.py
WRFitch: I'd argue that because this string has to be formatted this exact way every time on both charms, extracting it into its own function is necessary. This way, it can also can be tested independently of any other charm function.

tests/unit/test_pgb.py
WRFitch: As Jon suggested above, I've replaced this with a string constant that is formatted during the test.

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def generate_password() -> str:
+    """Generates a secure password of alphanumeric characters.
+
+    Passwords are alphanumeric only, to ensure compatibility with the userlist.txt format -
+    specifically, spaces and double quotes may interfere with parsing this file.
+
+    Returns:
+        A random 24-character string of letters and numbers.
```
quiet-ranger, line 47: Is the choice of only letters or numbers coming from a requirements document?  Is there a reason why we do not include punctuation characters, for example?   A simple clarifying comment is enough. If in doubt, we can ask around to see if there are any specific guidelines on this.

My objective here is not to be picky but to avoid it being bounced back further down the line.  Remember that the code is fully open sourced.

quiet-ranger (changes requested):

quiet-ranger:
As for the integration test failure, I see the following in the logs:

    raise JujuConnectionError('Model not found: {}'.format(model_name))
juju.errors.JujuConnectionError: Model not found: admin/itest

Could it be that it works locally because you created the model by hand but, in the integration tests, it does not exist yet?

tests/unit/test_pgb.py
```diff
@@ -0,0 +1,66 @@
...
+            listen_port=config["pgb_listen_port"],
+            listen_addr=config["pgb_listen_address"],
+            admin_users=",".join(users.keys()),
+        )
+        self.assertEqual(pgb_ini, expected_pgb_ini)
+
+    def test_generate_userlist(self):
+        users = {"test1": "pw1", "test2": "pw2"}
+        userlist = pgb.generate_userlist(users)
+        expected_userlist = '''"test1" "pw1"
```
MiaAltieri, line 40: is it possible to do `expected_userlist = '''"test1" "pw1"\n"test2" "pw2"'''` if so it would look nicer here :)

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -0,0 +1,105 @@
...
+        "juju-admin" "asdf1234"
+        '''
+    Returns:
+        users: a dictionary of usernames and passwords
+    """
+
+    parsed_userlist = {}
+    for line in userlist.split("\n"):
+        if line.strip() == "" or len(line.split(" ")) != 2:
+            continue
```
MiaAltieri, line 100: I'd add a `logger` message if we skip a line due to incorrect formatting of the line.

MiaAltieri (commented):
Looks _amazing_ just some knits

lib/charms/dp_pgbouncer_operator/v0/pgb.py
WRFitch: Good call, I've added a line of logging that states that if this is called, it's unable to parse a line in the userlist file. Unfortunately that line itself obviously can't be logged, since it contains a password, so it can't be too descriptive.

.github/workflows/ci.yaml
```diff
@@ -33,12 +37,13 @@ jobs:
...
+        # TODO set a predictable model name so it can be consumed by charm-logdump-action
+        # For now, setting one using "-- --model testing" causes it to get lost in the CI. This can be replicated locally.
+        run: tox -e integration
+      # - name: Dump logs
+      #   uses: canonical/charm-logdump-action@main
+      #   if: failure()
+      #   with:
+      #     # TEMPLATE-TODO: Replace the application name
+      #     app: operator-template
+      #     model: testing
```
WRFitch, line 49: To fix/replicate the integration testing issue, I've had to disable the part of the integration testing command that sets a model name, making this section unusable. However, there will also be a lot of changes to this section in future since this isn't a k8s charm and therefore shouldn't use microk8s for its integration testing. Therefore, I've left this content commented in and will update it in the next PR that provides a basic implementation for this charm.

quiet-ranger (approved):
Well done! To me this looks ready to be merged

tests/unit/test_pgb.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(pgb_ini, expected_pgb_ini)
+
+    def test_generate_userlist(self):
+        users = {"test1": "pw1", "test2": "pw2"}
+        userlist = pgb.generate_userlist(users)
+        expected_userlist = '''"test1" "pw1"\n"test2" "pw2"'''
+        self.assertEqual(userlist, expected_userlist)
+
+    def test_parse_userlist(self):
+        userlist = inspect.cleandoc(
```
MiaAltieri, line 46: I see what you're saying about keeping it as a constant, since its so test specific and I agree. Would be possible to put these contents in a file `tests/unit/data/userlist.txt` and then do: 

```
        with open("tests/unit/data/userlist.txt") as f:
            userlist = f.read()
```

MiaAltieri (commented):
Looks great! One suggestion

tests/unit/test_pgb.py
MiaAltieri, line 46: I see what you're saying about keeping it as a constant, since its so test specific and I agree. Would be possible to put these contents in a file `tests/unit/data/userlist.txt` and then do: 

```
        with open("tests/unit/data/userlist.txt") as f:
            userlist = f.read()
```

paulomach (approved):
LGTM

tests/unit/test_pgb.py
WRFitch: Thanks, I've added this in

PR #2, opened 2022-03-02
DPE-105 Adding pgbouncer.ini Config Management. 

WRFitch:
[Jira Ticket](https://warthogs.atlassian.net/browse/DPE-105?atlOrigin=eyJpIjoiZmVkZjk3MTcxYzM0NGJkMDg1NTU0NTQ5YjY1YWVlNWIiLCJwIjoiaiJ9)

This PR implements a class in the pgb charm library that can handle pgbouncer.ini config files, including parsing, editing, and writing. The existing implementation is a placeholder that implements the bare minimum for the pgbouncer application to come online, whereas the proposed implementation is capable of making use of the entire [pgbouncer config spec](https://www.pgbouncer.org/config.html). 

This class implements a mutablemapping that's copied from a [ConfigParser](https://docs.python.org/3/library/configparser.html) object, allowing config options to be accessed and modified dictionary-style, as well as providing some basic validation. The reason this doesn't just inherit from a ConfigParser is that it only uses strings, and therefore can't represent some of the config values that are better represented as dicts or lists.

Changelog: 
- Added PgbIni class
  - Added the ability to parse strings that represent pgbouncer.ini files
  - Added the ability to pass in pgbouncer config as a dictionary 
- Replaced placeholder pgbouncer.ini generation

### DIFF
--- a/lib/charms/dp_pgbouncer_operator/v0/pgb.py
+++ b/lib/charms/dp_pgbouncer_operator/v0/pgb.py
@@ -22,8 +22,9 @@ import io
 import logging
 import secrets
 import string
+from collections.abc import MutableMapping
 from configparser import ConfigParser
-from typing import Dict
+from typing import Dict, List
 
 logger = logging.getLogger(__name__)
 
@@ -41,19 +42,17 @@ def generate_password() -> str:
     return "".join([secrets.choice(choices) for _ in range(24)])
 
 
-def generate_pgbouncer_ini(dbconfig: Dict[str, Dict[str, str]], **kwargs) -> str:
+def generate_pgbouncer_ini(**kwargs) -> str:
     """Generate pgbouncer.ini file from the given config options.
 
     Args:
-        dbconfig: database config dictionary
         kwargs: kwargs following the [pgbouncer config spec](https://pgbouncer.org/config.html).
             Note that admin_users and stats_users must be passed in as lists of strings, not in
             their string representation in the .ini file.
     """
-    parser = IniParser()
-    cfg_dict = {"databases": dbconfig, "pgbouncer": kwargs}
-    parser.read_dict(cfg_dict)
-    return parser.write_to_string()
+    ini = PgbIni()
+    ini.parse_dict(kwargs)
+    return ini.render()
 
 
 def generate_userlist(users: Dict[str, str]) -> str:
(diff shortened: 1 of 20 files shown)


lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -96,15 +79,230 @@ def parse_userlist(userlist: str) -> Dict[str, str]:
         "juju-admin" "asdf1234"
         '''
     Returns:
-        users: a dictionary of usernames and passwords
+        users: a dictionary of valid usernames and passwords
     """
     parsed_userlist = {}
     for line in userlist.split("\n"):
-        if line.strip() == "" or len(line.split(" ")) != 2:
+        if (
```
paulomach, line 86: @WRFitch here you are testing for what can go wrong on the line, but wouldn't it be easier to test just what's not the exactly right criteria?

WRFitch: That's a good point, I've changed this to match against a regex string.

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -96,15 +79,230 @@ def parse_userlist(userlist: str) -> Dict[str, str]:
...
+        """Checks string is a valid database name.
+
+        For a database name to be valid, it must contain only alphanumeric characters, hyphens,
+        and underscores. Any other invalid character must be in double quotes.
+
+        Args:
+            string: the string to be validated
+        """
+        # Check dbnames are valid characters (alphanumeric and _- )
+        search = re.compile(r"[^A-Za-z0-9-_]+").search
```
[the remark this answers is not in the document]
zmraul: Huh, that's weird :) Thanks for the clarification!

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -96,15 +79,232 @@ def parse_userlist(userlist: str) -> Dict[str, str]:
...
+            string: the string to be validated
+        """
+        # Check dbnames don't use the reserved "pgbouncer" database name
+        if string == "pgbouncer":
+            raise PgbIni.IniParsingError(source=string)
+
+        # Check dbnames are valid characters (alphanumeric and _- )
+        search = re.compile(r"[^A-Za-z0-9-_]+").search
+        filtered_string = "".join(filter(search, string))
+        if len(filtered_string) == 0:
```
MiaAltieri, line 293: clever

MiaAltieri (commented):
Looking great 😎

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -96,15 +79,232 @@ def parse_userlist(userlist: str) -> Dict[str, str]:
...
             continue
-        # Userlist is formatted "{username}" "{password}""
+        # Userlist is formatted '"username" "password"'
         username, password = line.replace('"', "").split(" ")
         parsed_userlist[username] = password
 
     return parsed_userlist
+
+
+class PgbIni(MutableMapping):
```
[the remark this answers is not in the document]
WRFitch: PgbConfig is a better name, I'll update it to use that. Thanks!

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -96,15 +79,232 @@ def parse_userlist(userlist: str) -> Dict[str, str]:
...
+        if not set(essentials.keys()).issubset(set(self.keys())):
+            raise KeyError("necessary sections not found in config.")
+
+        if not set(essentials["pgbouncer"]).issubset(set(self["pgbouncer"].keys())):
+            raise KeyError("necessary pgbouncer config values not found in config.")
+
+        # Guarantee db names are valid
+        for db_id in self[db].keys():
+            db_name = self[db][db_id].get("dbname", "")
+            self._validate_dbname(db_id)
```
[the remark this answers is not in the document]
WRFitch: If the db_name parameter doesn't exist, pgbouncer will use the db_id as db_name, so they both have to be valid database names.

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -96,15 +79,232 @@ def parse_userlist(userlist: str) -> Dict[str, str]:
...
+
+        # No error checking for [databases] section, since it has to exist for pgbouncer to run.
+        for name, cfg_string in self[db].items():
+            self[db][name] = self._parse_string_to_dict(cfg_string)
+
+        try:
+            for name, cfg_string in self[users].items():
+                self[users][name] = self._parse_string_to_dict(cfg_string)
+        except KeyError:
+            # [users] section is not compulsory, so continue.
```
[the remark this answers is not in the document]
WRFitch: I would prefer to avoid this if possible - I don't want to unnecessarily log config information unless it's totally necessary, and this is just for skipping config values that don't exist. When we're using these values in the charm, I'll add logging to make sure anything that changes from a users' perspective is visible.

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -96,15 +79,232 @@ def parse_userlist(userlist: str) -> Dict[str, str]:
...
+            pass
+
+        for pgb_lst in PgbIni.pgb_list_entries:
+            try:
+                self[pgb][pgb_lst] = self[pgb][pgb_lst].split(",")
+            except KeyError:
+                # user fields are not compulsory, so continue
```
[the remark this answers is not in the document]
WRFitch: This has been updated to describe `list` fields, not `user` fields.

lib/charms/dp_pgbouncer_operator/v0/pgb.py
WRFitch: Thanks! :)

tests/unit/test_pgb.py
```diff
@@ -45,17 +60,117 @@ def test_generate_userlist(self):
...
+
+    def test_pgb_config_render(self):
+        with open(TEST_VALID_INI, "r") as test_ini:
+            input_string = test_ini.read()
+        output = pgb.PgbConfig(input_string).render()
+        self.assertEqual(input_string, output)
+
+    def test_pgb_config_validate(self):
+        # PgbConfig.validate() is called in read_string() and read_dict() methods, which are called
+        # in the constructor.
```
zmraul, line 134: Just a thought. If the error cases grow more, you could create a folder /data/error_cases/ or similar and move the pertinent files in there, so you can just loop trough all the files inside instead of having to load them 'by hand'.

zmraul (approved):
I think this is good! :+1:

tests/unit/test_pgb.py
WRFitch: This is a good idea for the future, I may need to change it based on expected errors but I'll definitely keep it in mind!

PR #3, opened 2022-03-09
DPE-107 Add Pr Template 

WRFitch:
## Issue
This repo has no PR templates, making it easy for PR authors to forget necessary information such as context or testing. 

[JIRA ticket](https://warthogs.atlassian.net/browse/DPE-107?atlOrigin=eyJpIjoiYWUyMmJiNTZiNTc2NDI1NGE5ZWYzNzI1YmMxNTdlMTAiLCJwIjoiaiJ9)

## Solution
Add a PR template that adds useful sections to PRs moving forward. 

## Context
This template differs slightly from [Shayan's proposed template](https://github.com/canonical/mysql-k8s-operator/pull/70). My reasoning for changing the template in this repo without propagating my feedback to the original PR is as follows:
- I want to start using PR templates as soon as possible. It's easy to endlessly discuss process, but we won't fully understand what's useful information and what's a chore until we try it. 
- I'd prefer for these standards to be more flexible - for example, a testing section is necessary for code changes, but unnecessary for documentation changes. I'd like for PR authors to change these templates at their discretion. 

The plan, therefore, is to implement a minimal template that doesn't enforce box-ticking and unnecessary explanations, while pushing PR authors to include information necessary for reviewers.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,19 +1,14 @@
-# Issue
+## Issue
 <!-- What issue is this PR trying to solve? -->
 
-
-# Solution
+## Solution
 <!-- A summary of the solution addressing the above issue. -->
 
-
-# Context
+## Context
 <!-- Necessary details to understand the proposed changes. -->
 
-
-# Release Notes
+## Release Notes
 <!-- A digestable summary of the changes in this PR. -->
 
-
-# Testing
+## Testing
 <!-- A summary of how this PR has been tested. -->
-
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,19 @@
+# Issue
+<!-- What issue is this PR trying to solve? -->
+
+
+# Solution
+<!-- A summary of the solution addressing the above issue. -->
+
+
+# Context
+<!-- Necessary details to understand the proposed changes. -->
+
+
+# Release Notes
+<!-- A digestable summary of the changes in this PR. -->
+
+
+# Testing
+<!-- A summary of how this PR has been tested. -->
+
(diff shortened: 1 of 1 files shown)


shayancanonical (approved):
lgtm! as you said, we can always delete the non-relevant portions of the PR template when creating a PR

paulomach (approved):
I like it. I feel the solution part should be optional, since the commits are the solution. But it's good for starter. 
My thoughts are that we should promote this (or some form) to the operator template repository. @jnsgruk ?

zmraul:
> I like it. I feel the solution part should be optional, since the commits are the solution. But it's good for starter. My thoughts are that we should promote this (or some form) to the operator template repository. @jnsgruk ?

I would suggest making a PR to the operator template like this one, so everyone interested can comment. Right now is a bit hidden in here :)

zeljkoagic (approved):

zeljkoagic:
> I feel the solution part should be optional, since the commits are the solution.

I see where this is coming from, but it's still good to have a two-liner explanation of the approach taken, as otherwise we run the risk of subscribing to "the code is the documentation" paradigm. ;-)

WRFitch:
> I would suggest making a PR to the operator template like this one, so everyone interested can comment. Right now is a bit hidden in here :)

I'd prefer to test the templates before setting them as a standard for all new operator repos. Once we've used them in a few PRs and we're confident that they're useful in their current form, I'm happy to add this into the operator-template repo.

zmraul:
> I'd prefer to test the templates before setting them as a standard for all new operator repos. Once we've used them in a few PRs and we're confident that they're useful in their current form, I'm happy to add this into the operator-template repo.

Sure! Maybe in the form of an issue or similar. I meant it so the discussion on how to do something like this is not siloed here or on random PR across our repos.

PR #4, opened 2022-03-15
Dpe 17 pgb handle install event

WRFitch:


actions.yaml
```diff
@@ -1,16 +1,3 @@
 # Copyright 2022 Canonical Ltd.
```
shayancanonical, line 1: should we be committing an empty actions file? is it required by charmcraft while packing?

metadata.yaml
```diff
@@ -1,25 +1,17 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-# For a complete list of supported options, see:
-# https://discourse.charmhub.io/t/charm-metadata-v2/3674/15
-name: operator-template
+name: dp-pgbouncer-operator
```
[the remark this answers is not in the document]
WRFitch: It was a suggested guideline, at least while we're taking over from existing charms in charmhub. I'm happy to remove it if nobody else is following the standard, though. 

DP stands for Data Platform, yes.

lib/charms/dp_pgbouncer_operator/v0/pgb.py
```diff
@@ -232,7 +168,7 @@ def render(self) -> str:
         self.validate()
```
shayancanonical, line 168: nit: is `validate()` a private method? if so, should it be named `_validate()`?

metadata.yaml
```diff
@@ -1,25 +1,17 @@
...
-resources:
-  httpbin-image:
-    type: oci-image
-    description: OCI image for httpbin (kennethreitz/httpbin)
-    # Included for simplicity in integration tests
-    upstream-source: kennethreitz/httpbin
+series:
+  # TODO add jammy once it's available
+  - focal
+  - bionic
```
shayancanonical, line 17: nit: new line

lib/charms/dp_pgbouncer_operator/v0/pgb.py
WRFitch: It isn't intended as a private method. The idea is that when you're changing the config, you can check that the changes you've made are valid and reject those changes. For example, if a user proposed a config change that introduces an invalid database name, I think it's preferable to reject that when it's proposed by the user, rather than when we try to write the invalid config.

src/charm.py
```diff
@@ -1,104 +1,195 @@
...
+        except subprocess.CalledProcessError as e:
+            logger.exception("failed to update apt cache, CalledProcessError", exc_info=e)
+            self.unit.status = BlockedStatus("failed to update apt cache")
+            return
+
+        try:
+            logger.debug(f"installing apt packages: {', '.join(packages)}")
+            apt.add_package(packages)
+        except apt.PackageNotFoundError:
+            logger.error("a specified package not found in package cache or on system")
```
shayancanonical, line 118: is it possible to extract the package name that was not found in the package cache?

actions.yaml
WRFitch: It's not required by charmcraft, I just forgot to remove it when deleting the boilerplate actions

src/charm.py
```diff
@@ -1,104 +1,195 @@
...
+            self.unit.status = BlockedStatus("failed to install packages")
+
+    def _render_file(self, path: str, content: str, mode: int) -> None:
+        """Write a content rendered from a template to a file.
```
shayancanonical, line 122: very minor nit: `write content`

metadata.yaml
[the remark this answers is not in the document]
paulomach: Not sure about that: it's the first time I saw it.
Personally I don't like it, since is makes no sense from the canonical outsider perspective and we are not publishing to CH still, so  there's no clash.

src/charm.py
```diff
@@ -1,104 +1,195 @@
...
 
     _stored = StoredState()
 
     def __init__(self, *args):
         super().__init__(*args)
-        self.framework.observe(self.on.httpbin_pebble_ready, self._on_httpbin_pebble_ready)
+
+        self._pgbouncer_service = "pgbouncer"
+        self._pgb_user = "pgbouncer"
+        self._pgb_uid = 1100
```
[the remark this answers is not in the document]
WRFitch: It's possible to deploy multiple charms on one machine, and that's part of the intended use case for PgBouncer, so this is a valid concern. I'm going to try generating the user and then access the uid/gid using `pwd` similar to the `_render_file` method, thanks for picking this up.

src/charm.py
WRFitch: Thanks for picking this up!

tests/unit/test_charm.py
```diff
@@ -1,67 +1,167 @@
...
-        self.assertTrue(service.is_running())
-        # Ensure we set an ActiveStatus with no message
-        self.assertEqual(self.harness.model.unit.status, ActiveStatus())
+    @patch("charms.operator_libs_linux.v0.passwd.add_user")
+    @patch("charms.dp_pgbouncer_operator.v0.pgb.initialise_userlist_from_ini")
+    @patch("charm.PgBouncerCharm._install_apt_packages")
+    @patch("charm.PgBouncerCharm._render_file")
+    @patch("os.setuid")
+    @patch("os.chown")
+    def test_on_install(self, _chown, _setuid, _render_file, _install, _userlist, _add_user):
```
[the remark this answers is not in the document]
WRFitch: This already exists at line 87 :)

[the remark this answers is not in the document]
shayancanonical: Oops, totally missed it

paulomach (approved):
Looks good to go!

src/charm.py
```diff
@@ -1,104 +1,198 @@
...
-                    "command": "gunicorn -b 0.0.0.0:80 httpbin:app -k gevent",
-                    "startup": "enabled",
-                    "environment": {"thing": self.model.config["thing"]},
-                }
+        self.unit.status = MaintenanceStatus("Installing and configuring PgBouncer")
+
+        # Initialise prereqs to run pgbouncer
+        self._install_apt_packages(["pgbouncer"])
+        # add pgbouncer user to postgres group, created in above line
+        passwd.add_user(
```
paulomach, line 56: @WRFitch , one idea appeared in my head still around this subject. 
It's possible that this is an install event on a machine where the event already happened.
If so, unless the charm removal take care the user deletion, you should test if the user already exists. 
I'm not aware if this is already covered by the function in the library, if it is, dismiss the whole comment!

src/charm.py
```diff
@@ -1,104 +1,198 @@
...
+            username=self._pgb_user, password="pgb", primary_group="postgres"
```
marceloneppel, line 57: Probably is better to have a generated password here too.

src/charm.py
```diff
@@ -1,104 +1,198 @@
...
+
+        # Initialise userlist, generating passwords for initial users
+        self._render_userlist(pgb.initialise_userlist_from_ini(ini))
+
+        self.unit.status = WaitingStatus("Waiting to start PgBouncer")
+
+    def _on_start(self, _) -> None:
+        try:
+            # -d flag ensures pgbouncer runs as a daemon, not as an active process.
+            command = ["pgbouncer", "-d", INI_PATH]
```
marceloneppel, line 90: This probably can be replaced by the use of `systemd` charm library call. But it can be evaluated and implemented in another PR.

tests/unit/test_pgb.py
```diff
@@ -174,3 +112,88 @@ def test_pgb_config__validate_dbname(self):
...
+            "another_testuser": "anotherpass",
+            "1234": "",
+            "": "",
+        }
+        self.assertDictEqual(users, expected_users)
+
+        # Check that we can run input through a few times without anything getting corrupted.
+        regen_userlist = pgb.generate_userlist(users)
+        regen_users = pgb.parse_userlist(regen_userlist)
+        self.assertNotEqual(regen_userlist, userlist)
```
marceloneppel, line 198: Maybe a comment here can make more clear in the test code why the user lists are different (because there are invalid values in `userlist` content).

marceloneppel (changes requested):
Great work, Will! I indicated a change in the docs and left a few other comments.

src/charm.py
jnsgruk: Yeh, for sure this process should be being managed by an appropriate systemd unit file - but as Marcelo said, this can be in a separate PR.

metadata.yaml
```diff
@@ -1,25 +1,17 @@
...
-    resource: httpbin-image
-
-# TEMPLATE-TODO: each container defined above must specify an oci-image resource
-resources:
-  httpbin-image:
-    type: oci-image
-    description: OCI image for httpbin (kennethreitz/httpbin)
-    # Included for simplicity in integration tests
-    upstream-source: kennethreitz/httpbin
+series:
```
jnsgruk, line 14: Series is deprecated in favour of the `bases` specified in `charmcraft.yaml` - you can remove this.

metadata.yaml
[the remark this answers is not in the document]
WRFitch: This has been removed - it's an awkward name, and the dp- prefix isn't a standard if it's only used in one place.

CONTRIBUTING.md
```diff
@@ -50,14 +48,16 @@ charmcraft pack
...
 ```bash
 # Create a model
 juju add-model dev
 # Enable DEBUG logging
 juju model-config logging-config="<root>=INFO;unit=DEBUG"
 # Deploy the charm
-juju deploy ./template-operator_ubuntu-20.04-amd64.charm \
-    --resource httpbin-image=kennethreitz/httpbin \
+juju deploy ./pgbouncer-operator_ubuntu-20.04-amd64.charm \
+    --resource pgbouncer-image=pgbouncer:0.8
```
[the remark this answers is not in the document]
WRFitch: Thanks for catching this, it's been removed

src/charm.py
WRFitch: I'm intending to add better service management (health, logging, etc) in a later PR, so I'll add your feedback into the jira ticket.

src/charm.py
WRFitch: This is definitely a valid concern with one intended use case for this charm, but fortunately the [passwd library](https://github.com/canonical/operator-libs-linux/blob/2020c04f38eb8d6b5569a18c40b416d5a20a298b/lib/charms/operator_libs_linux/v0/passwd.py#L93) already checks for this.

marceloneppel (approved):

PR #5, opened 2022-03-29
DPE-136 Added Charm Config

WRFitch:
## Issue
The charm has to expose some config options for pgbouncer to be configurable. This draft PR exists so we can discuss what is necessary and what is unnecessary, out of the proposed changes and the [pgbouncer config docs](https://www.pgbouncer.org/config.html)

Jira Ticket: [DPE-136](https://warthogs.atlassian.net/browse/DPE-136?atlOrigin=eyJpIjoiYmE3ZmJmMDJhODU0NGNlYjkyNmYxYTllYTk4NTJhMjMiLCJwIjoiaiJ9)

## Solution
This PR proposes a minimal config, whose options should provide enough information for the charm to calculate the rest of the config values that relate to pooling. The rest of the pgbouncer config should be defined as best practices by the charm, such as users and TLS settings. 

## Context
The defaults and documentation for these values are copied wholesale from the pgbouncer config docs. If this is unacceptable, please let me know so it can be changed. 

## Release Notes
- Adds a subset of pgbouncer config values to charm config.


config.yaml
```diff
@@ -2,17 +2,682 @@
...
-    default: "localhost"
-    description: "Address on which PGBouncer listens for traffic"
-    type: "string"
+  # Charm Settings
+  reload_on_config_changed:
+    default: true
+    description: |
+      Whether to automatically reload pgbouncer after every config change, or
+      rely on manual reloads using the reload-pgbouncer action.
+    type: boolean
```
jnsgruk, line 12: This feels like a thing that should be decided by the charm... I don't think this should be included.

config.yaml
```diff
@@ -2,17 +2,682 @@
...
+
+      Direct access to pg_shadow requires admin rights. It’s preferable to use
+      a non-superuser that calls a SECURITY DEFINER function instead.
+    type: string
+
+  # Log Settings
+  syslog:
+    default: 0
+    description: Toggles syslog on/off.
+    type: int
```
jnsgruk, line 217: I'm not super clear what this means - if disabled, does this mean pgbouncer just logs to stdout and doesn't retain log files?

config.yaml
```diff
@@ -2,17 +2,682 @@
...
+      disabled if external monitoring tools are used to grab the same data from
+      SHOW commands.
+    type: int
+
+  verbose:
+    default: 0
+    description: |
+      Increase verbosity. Mirrors the “-v” switch on the command line. For
+      example, using “-v -v” on the command line is the same as verbose=2.
+    type: int
```
jnsgruk, line 247: I think these could be combined into a single option that has 2 or 3 different profiles.... charm config doesn't need to be a 1:1 mapping with workload config :)

config.yaml
```diff
@@ -2,17 +2,682 @@
...
+      PgBouncer can collect DNS zones from host names (everything after first
+      dot) and then periodically check if the zone serial changes. If it
+      notices changes, all host names under that zone are looked up again. If
+      any host IP changes, its connections are invalidated.
+
+      Works only with UDNS and c-ares backends (configure option --with-udns
+      or --with-cares).
+    type: float
+
+  # TLS Settings
```
jnsgruk, line 394: In general, it feels like we should aim for a more elegant solution for TLS that utilises realtions. This feels very error prone to me.

config.yaml
WRFitch: Pgbouncer logs to stdout unless it is run as a daemon, but it also logs to wherever the `logfile` config parameter points to. From some googling, it looks like this exists so you can use eventlog on Windows, making this value irrelevant for a charm.

config.yaml
WRFitch: I've replaced these with the following, let me know what you think: 
```yaml
  # Log Settings
  log_level:
    default: normal
    description: |
      The level to which information is logged. Similar to the pgbouncer
      "verbose" option, but also handles connections, disconnections, pooler
      errors, and statistics. Controls the following pgbouncer options:
      "log_connections", "log_disconnections", "log_pooler_errors",
      "log_stats", "verbose".

      Levels:
      stats: verbose 0, with statistics.
      normal: verbose 1, logs connection information and pooler errors.
      full: max verbose, logs connection information, pooler errors, and 
      statistics.
    type: string
```

jnsgruk: Much better :)

jnsgruk (approved):

marceloneppel (approved):
LGTM!

README.md
```diff
@@ -6,11 +6,45 @@ The PgBouncer Operator deploys and operates the [PgBouncer](https://www.pgbounce
...
+
+- effective_db_connections = max_db_connections / number of charm instances
+- default_pool_size = effective_db_connections / 2
+- min_pool_size = effective_db_connections / 4
+- reserve_pool_size = effective_db_connections / 4
+- ignore_startup_parameters = extra_float_digits (extracted from postgres charm)
+
+The following config values are set as constants in the charm:
+
+- max_client_conn = 10000
```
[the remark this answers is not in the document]
WRFitch: This isn't set in the charm - I'm going to be adding all this functionality in the next PR, this PR is just for us to argue over what config we want.

PR #6, opened 2022-04-08
Dpe 21 handle config change

WRFitch:
## Issue
Charm config changes need to be applied to the pgbouncer application through pgbouncer.ini. This includes generating some variables for.

## Release Notes
- Add default config to pgb library
- Add config-changed handler
- Added method to generate pgbouncer config values from charm config. 
- Documentation & QOL updates

## Testing
- Manual testing
  - Set config, check file is updated correctly
- Unit testing
- Integration testing
  - Set config, check file is updated correctly


lib/charms/pgbouncer_operator/v0/pgb.py
```diff
@@ -239,6 +257,32 @@ def _validate_dbname(self, string: str):
...
+            raise PgbConfig.ConfigParsingError(source="pgb_instances cannot be less than 1 ")
+
+        pgb = PgbConfig.pgb_section
+
+        self[pgb]["max_db_connections"] = str(max_db_connections)
+
+        effective_db_connections = max_db_connections / pgb_instances
+        self[pgb]["default_pool_size"] = str(int(effective_db_connections / 2))
+        self[pgb]["min_pool_size"] = str(int(effective_db_connections / 4))
+        self[pgb]["reserve_pool_size"] = str(int(effective_db_connections / 4))
```
marcoppenheimer, line 284: From a PGBouncer perspective (obviously I have no knowledge on this), what's the logic behind settings these vars with these fractions? Do `min_pool_size` and `reserve_pool_size` need to be equal always?
Does `min_pool_size` need to be half of `default_pool_size` always?

Or am I just lost and dividing by 2 and by 4 is the 'logic' I'm talking about?

I would try and make these items more explicit to convey the logic behind them.

marcoppenheimer: I also presume that the `int()` here is really to `floor` non-integer results of the fractions, is that right? If so, I'd say `math.floor()`, just so it's clear.

tests/unit/test_pgb.py
```diff
@@ -101,17 +101,34 @@ def test_pgb_config_validate(self):
...
+        self.assertEqual(cfg["pgbouncer"]["max_db_connections"], str(max_db_connections))
+
+        effective_db_connections = max_db_connections / pgb_instances
+        self.assertEqual(
+            cfg["pgbouncer"]["default_pool_size"], str(int(effective_db_connections / 2))
+        )
+        self.assertEqual(cfg["pgbouncer"]["min_pool_size"], str(int(effective_db_connections / 4)))
+        self.assertEqual(
+            cfg["pgbouncer"]["reserve_pool_size"], str(int(effective_db_connections / 4))
+        )
```
marcoppenheimer, line 131: These are redundant as they just map to the logic from the code. 
Maybe try and generalise given knowledge of what PGBouncer needs as values for these config items, similar to my comment above.

Something like:
```
min_pool_size == reserve_pool_size
default_pool_size == 11
etc
```

marcoppenheimer (approved):
Looks good, I added loads of comments mostly around testing + structure, but the core is solid! 💪🏽

src/charm.py
```diff
@@ -96,11 +90,24 @@ def _on_start(self, _) -> None:
...
-        pass
+        """Config changed handler.
+
+        Reads config values and parses them to pgbouncer config, restarting if necessary.
+        """
+        config = self._read_pgb_config()
+        config["pgbouncer"]["pool_mode"] = self.config["pool_mode"]
+
+        config.set_max_db_connection_derivatives(
+            self.config["max_db_connections"],
```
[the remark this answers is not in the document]
marcoppenheimer: I can't remember what `self.config.get()` returns off the top of my head, I just recall that the items are all stored as strings, so don't know if they convert when returning it to you.

lib/charms/pgbouncer_operator/v0/pgb.py
WRFitch: Setting these configs this way is more a rule of thumb, but I'll add some documentation explaining why these are set this way because it currently looks pretty arbitrary. Some explanation for now:
- We have to split the total number of connections across local pgbouncer instances, and the easiest way to do that is splitting it equally. 
- Default values for the pool sizes are usually smaller in the [Pgbouncer config docs](https://www.pgbouncer.org/config.html), but  larger defaults, mins, and reserves means we can make better use of how flexible a cluster is:
  - larger `default_pool_size` means each new unit will have plenty of spare space when it comes online
  - larger `min_pool_size` and `reserve_pool_size` means when a unit goes down, the other units will have plenty of space to take on its workload. 
- `min_pool_size` and `reserve pool_size` don't always have to be equal

`int()` is only here because we can't have a fraction of a connection. I suppose `math.ceil()` would be more useful though, since setting a value to 0 can often mean unlimited connections in pgbouncer, so rounding down is slightly dangerous.

src/charm.py
```diff
@@ -86,6 +76,10 @@ def _on_install(self, _) -> None:
         self.unit.status = WaitingStatus("Waiting to start PgBouncer")
 
     def _on_start(self, _) -> None:
+        """On Start hook.
+
+        Switches to pgbouncer user and runs pgbouncer as daemon
+        """
```
marceloneppel, line 82: I saw that there is also another process running with user postgres. Using the top command inside the unit, I could see two processes running:

```
PID  USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
4673 postgres  20   0   15752   2412   1644 S   0.0   0.0   0:00.18 /usr/sbin/pgbouncer -d /etc/pgbouncer/pgbouncer.ini
5088 pgbounc+  20   0   15752   2428   1668 S   0.0   0.0   0:00.13 pgbouncer -d /etc/pgbouncer/pgbouncer.ini
```

Is it expected to have two processes at this moment? And with different users? Maybe it was started automatically when installing pgbouncer through `apt`. PostgreSQL has the same behavior.

marceloneppel (approved):
LGTM! I left just one comment about the processes that are started in the unit.

src/charm.py
```diff
@@ -96,11 +90,24 @@ def _on_start(self, _) -> None:
...
+
+        Reads config values and parses them to pgbouncer config, restarting if necessary.
+        """
+        config = self._read_pgb_config()
+        config["pgbouncer"]["pool_mode"] = self.config["pool_mode"]
+
+        config.set_max_db_connection_derivatives(
+            self.config["max_db_connections"],
+            os.cpu_count(),
+        )
```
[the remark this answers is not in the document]
WRFitch: As we discussed, since I intended the PgbConfig object to be a python representation of the pgbouncer.ini config, this suggestion slightly distorts that abstraction towards representing the pgbouncer application and how it's deployed. However, you're right that the code in charm.py should be simpler, so in future I'm going to experiment with a Pgbouncer object in the charm library, which should abstract this logic away from the charm.

tests/unit/test_charm.py
```diff
@@ -86,6 +85,33 @@ def test_on_start(self, _call, _getpwnam, _setuid):
...
+                "max_db_connections": max_db_connections,
+            }
+        )
+
+        _read.assert_called_once()
+        # _read.return_value is modified on config update, but the object reference is the same.
+        _render.assert_called_with(_read.return_value, reload_pgbouncer=True)
+
+        self.assertDictEqual(dict(_read.return_value), dict(config))
+
```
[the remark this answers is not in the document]
WRFitch: I think in general you're right, but it also tests that the pgbouncer application is restarted (line 111, `reload_pgbouncer=True), which is critical for applying changes to the running pgbouncer application from the config file. There already is an integration test that checks what you've asked, which does make this test mostly superfluous, but it's easier to check that `PgBouncerCharm._reload_pgbouncer()` is definitely called this way.

src/charm.py
WRFitch: This is not expected behaviour, I'll look into this.

src/charm.py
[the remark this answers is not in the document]
WRFitch: Not exactly - because its type is listed as an `int` in config.yaml, it returns an int.

lib/charms/pgbouncer_operator/v0/pgb.py
WRFitch: This has been documented in README.md and config.yaml.

paulomach (approved):
Mint

tests/unit/test_charm.py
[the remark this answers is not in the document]
marcoppenheimer: This is much better 👍🏽 💪🏽

tests/unit/test_pgb.py
```diff
@@ -101,17 +101,26 @@ def test_pgb_config_validate(self):
...
+                config._validate_dbname(dbname)
+
+    def test_set_max_db_connection_derivatives(self):
+        cfg = pgb.PgbConfig(pgb.DEFAULT_CONFIG)
+        cfg.set_max_db_connection_derivatives(44, 2)
+
+        self.assertEqual(cfg["pgbouncer"]["max_db_connections"], "44")
+        self.assertEqual(cfg["pgbouncer"]["default_pool_size"], "11")
+        self.assertEqual(cfg["pgbouncer"]["min_pool_size"], "6")
+        self.assertEqual(cfg["pgbouncer"]["reserve_pool_size"], "6")
```
marcoppenheimer, line 123: Super small change, would suggest adding a test based on what you mentioned in the discussions above, to ensure that they never return `0` for the fields where that might be a problem.

So I dunno, `cfg.set_max_db_connection_derivatives(3,1)` expects `3, 2, 1, 1`, or whatever.

marcoppenheimer: Only adding as you said it might cause issues, sounds like the right thing to test :)

marcoppenheimer (approved):
Looks great!

tests/unit/test_pgb.py
WRFitch: I spent some more time looking at this today, and the option to set max_db_connections=0 either needs to be blocked or handled directly with some sensible defaults that suit a charm deployment. I've opened [a jira ticket](https://warthogs.atlassian.net/browse/DPE-208?atlOrigin=eyJpIjoiNDQ3N2Y3MTcxOTdiNGU4MzgyM2Y2MjcxNDVhNzg1NTMiLCJwIjoiaiJ9) to handle this, since I haven't figured out what those defaults are yet.

WRFitch: I've implemented a solution based on the jira ticket, let me know what you think

tests/unit/test_pgb.py
```diff
@@ -101,17 +101,35 @@ def test_pgb_config_validate(self):
...
+
+        # Test defaults when max_db_connection is unlimited
+        cfg.set_max_db_connection_derivatives(0, 123252)
+
+        self.assertEqual(cfg["pgbouncer"]["max_db_connections"], "0")
+        self.assertEqual(cfg["pgbouncer"]["default_pool_size"], "20")
+        self.assertEqual(cfg["pgbouncer"]["min_pool_size"], "10")
+        self.assertEqual(cfg["pgbouncer"]["reserve_pool_size"], "10")
```
marceloneppel, line 132: Nice addition!

marceloneppel (approved):

lib/charms/pgbouncer_operator/v0/pgb.py
```diff
@@ -117,8 +130,11 @@ def _parse_complex_variables(self) -> None:
         pgb = PgbConfig.pgb_section
 
         # No error checking for [databases] section, since it has to exist for pgbouncer to run.
-        for name, cfg_string in self[db].items():
-            self[db][name] = self._parse_string_to_dict(cfg_string)
+        try:
+            for name, cfg_string in self[db].items():
+                self[db][name] = self._parse_string_to_dict(cfg_string)
+        except KeyError as err:
+            raise PgbConfig.ConfigParsingError(source=str(err))
```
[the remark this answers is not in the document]
WRFitch: Whoops, I missed that. I've removed the comment, since this obviously does check for the [databases] section.

marceloneppel (approved):

delgod (approved):

PR #7, opened 2022-04-20
DPE-122 Run PgBouncer through systemd

WRFitch:
## Proposal
Running pgbouncer using a subshell command prevents running multiple instances, and doesn't make use of systemd's features such as automatically restarting processes, and logging. This PR uses the operator_libs_linux systemd charm library to run and control pgbouncer. 

## Context
- pgbouncer is single-threaded, so we need to run multiple instances of it to make use of multiple cores. This will be implemented in a future PR. 
- Postgres user is added when pgbouncer is installed. 

## Release Notes
- add systemd charm library
  - There is no need to review `lib/charms/operator_libs_linux/v1/systemd.py`
- Removed pgbouncer user
- Moved pgbouncer config to postgres user home directory (`/var/lib/postgresql/pgbouncer`)
- Implemented systemd service file for pgbouncer
- Implemented pgbouncer service management through systemd
- Fixed a false positive test in charm library

## Testing
- Manually tested systemd restarts by killing pgbouncer process and observing it stays online
- Written unit tests for new functions
- Existing integration tests have not been modified, but one has been added to implement the above manual testing steps


src/charm.py
```diff
@@ -32,12 +35,13 @@ class PgBouncerCharm(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
 
-        self._pgbouncer_service = "pgbouncer"
-        self._pgb_user = "pgbouncer"
+        self._pgbouncer_service = PGB
```
shayancanonical, line 38: is `_pgbouncer_service` needed as a property on the charm class? can we instead just use PGB directly wherever we would call `self._pgbouncer_service`?

src/charm.py
```diff
@@ -51,50 +55,55 @@ def _on_install(self, _) -> None:
...
-        # pgbouncer apt package
-        passwd.add_user(
-            username=self._pgb_user, password=pgb.generate_password(), primary_group="postgres"
-        )
-        user = pwd.getpwnam(self._pgb_user)
+        self._install_apt_packages([self._pgbouncer_service])
+
+        user = pwd.getpwnam(self._postgres_user)
         self._postgres_gid = user.pw_gid
         self._pgbouncer_uid = user.pw_uid
```
shayancanonical, line 62: From what I see, `self._postgres_gid` and `self._pgbouncer_uid` are only used on line 65 when chown-ing the PGB_DIR. Would it make sense to make the them local variables instead of class properties?

src/charm.py
```diff
@@ -202,7 +211,14 @@ def _render_userlist(self, userlist: Dict[str, str], reload_pgbouncer: bool = Fa
...
 
     def _reload_pgbouncer(self):
-        self.unit.status = ActiveStatus("pgbouncer has reloaded")
+        """Reloads systemd pgbouncer service."""
+        self.unit.status = MaintenanceStatus("Reloading Pgbouncer")
+        try:
+            systemd.service_reload(PGB, restart_on_failure=True)
+            self.unit.status = ActiveStatus("PgBouncer Reloaded")
+        except systemd.SystemdError as e:
+            logger.info(e)
```
shayancanonical, line 220: use logger.exeception instead?

src/charm.py
```diff
@@ -32,12 +35,13 @@ class PgBouncerCharm(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
 
-        self._pgbouncer_service = "pgbouncer"
-        self._pgb_user = "pgbouncer"
+        self._pgbouncer_service = PGB
+        self._postgres_user = "postgres"
```
shayancanonical, line 39: should `_postgres_user` be a constant in the file (like PGB)?

src/pgbouncer.service
```diff
@@ -0,0 +1,14 @@
...
+[Service]
+Type=simple
+User=postgres
+ExecStart=/sbin/pgbouncer /var/lib/postgresql/pgbouncer/pgbouncer.ini
+ExecReload=/bin/kill -HUP $MAINPID
+KillSignal=SIGINT
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
```
shayancanonical, line 14: missing new line

src/charm.py
```diff
@@ -51,50 +55,55 @@ def _on_install(self, _) -> None:
...
         # Initialise pgbouncer.ini config file from defaults set in charm lib.
         ini = pgb.PgbConfig(pgb.DEFAULT_CONFIG)
         self._render_pgb_config(ini)
-
         # Initialise userlist, generating passwords for initial users
         self._render_userlist(pgb.initialise_userlist_from_ini(ini))
 
+        # Enable pgbouncer and reload systemd
+        shutil.copy("src/pgbouncer.service", "/etc/systemd/system/pgbouncer.service")
+        systemd.daemon_reload()
```
paulomach, line 75: Could no infer from [libs documentaion](https://charmhub.io/operator-libs-linux/libraries/systemd), but normally on systemd you need to enable the unit (in systemd terms) to allow it autostart (with `systemctl enable <service>`).
Are you always controlling startup in the charm or it's desirable to have systemd start it?

paulomach: Yes you are (saw it below).. 
But have you consider having the host systemd initialising it for you?

src/charm.py
```diff
@@ -19,7 +21,8 @@
 
 logger = logging.getLogger(__name__)
 
-PGB_DIR = "/etc/pgbouncer"
+PGB = "pgbouncer"
+PGB_DIR = "/var/lib/postgresql/pgbouncer"
```
zmraul, line 25: Would it make sense to have `PGB_DIR` value coming from the relation with postgresql?

WRFitch: `PGB_DIR` is a directory to contain config values that are unique to pgbouncer, some of which will be set based on the postgres relation. Since it exists for pgbouncer-specific purposes, it will be controlled by the pgbouncer charm.

src/charm.py
WRFitch: I've removed these variables entirely, since they were unnecessarily cluttered. Thanks for the cleanup suggestions!

src/charm.py
WRFitch: You're correct, this has been removed.

paulomach (approved):
LGTM

src/charm.py
WRFitch: Based on our discussion, while we want to make use of systemd as much as possible, I think it makes the charm code more readable to control daemon start/stop directly, rather than letting systemd autostart the service. I also think this is why the systemd charm lib only has an option to enable a service when starting it; the idea (as it appears to me) is to directly control starting and stopping, but allow restarts and service health to be handled by systemd.

marceloneppel (approved):
LGTM!

PR #9, opened 2022-06-14
Dpe 185 postgres relation

WRFitch:
## Proposal
This PR adds support for the legacy backend-db-admin relation to the postgresql charm. 

## Context
- This PR isn't currently ready for review, since I'm still testing it. 

## Release Notes
- Added support for legacy backend-db-admin relation

## Testing
- Integration tests
- manual testing
- unit tests are still being written


tests/integration/test_charm.py
```diff
@@ -83,3 +88,59 @@ async def test_systemd_restarts_pgbouncer_processes(ops_test: OpsTest):
...
     # verify all processes start again
     assert await helpers.get_running_instances(unit, "pgbouncer") == expected_processes
+
+
+# ===================================
+#  Pgbouncer-postgres relation tests
+# ===================================
+
+
+async def test_backend_db_admin_relation_slowtest(ops_test: OpsTest):
```
marceloneppel, line 98: Maybe we can move the tests related to the old relations to another file. I saw some comments about it in other charms.

tox.ini
```diff
@@ -73,4 +73,16 @@ deps =
...
+[testenv:fast-integration]
+description = run integration tests, skipping the slow ones.
+deps =
+    pytest
+    juju
+    pytest-operator
+    -r{toxinidir}/requirements.txt
+commands =
+    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs} \
+        --durations=0 -k "not slowtest"
```
marceloneppel, line 88: Good idea to have this :)

marceloneppel (commented):
It seems a great start!

tests/integration/test_charm.py
WRFitch: Yeah I also noticed John's comment about moving the legacy relation handlers into their own files - I'll give this a look later today

WRFitch: This has been done :)

requirements.txt
```diff
@@ -1 +1,2 @@
-ops >= 1.4.0
\ No newline at end of file
+ops >= 1.4.0
+pgconnstr >= 1.0.1
```
marceloneppel, line 2: You can remove this dependency as it's not being used.

marceloneppel (approved):
LGTM! Great work, Will! Part of this code will be useful for the new relation.

requirements.txt
WRFitch: Thanks for catching this!

src/relations/backend_db_admin.py
```diff
@@ -0,0 +1,162 @@
...
+        logger.info(
+            "DEPRECATION WARNING - backend-db-admin is a legacy relation, and will be deprecated in a future release. "
+        )
+
+        cfg = self.charm._read_pgb_config()
+        cfg["databases"].pop("pg_master", None)
+
+        # Get postgres leader unit from relation data through iteration. Using departed_event.unit
+        # appears to pick a unit at random, and relation data is not copied over to
+        # departed_event.app, so we do this instead.
```
MiaAltieri, line 149: I could be wrong on this or also mis-interpretting this comment, but I think if you have this set up:
```
juju status

App                  Version  Status  Scale  Charm        Channel  Rev  Exposed  Message
postgres           active      1  application             1  no       
pgbouncer              active      1  application                 0  no     

Unit                    Workload  Agent  Machine  Public address  Ports      Message
pgbouncer/0*  active    idle   5        10.209.221.142             
postgres/0*     active    idle   4        10.209.221.33  
postgres/1          active    idle   0        10.209.221.202           
postgres/2           active    idle   1        10.209.221.92              
```

and the `postgres` application departs from the `pgbouncer` application, the unit:
`pgbouncer` will get three `_on_relation_departed` events one for `postgres/0*`, `postgres/1`, `postgres/2` (in no particular order (ie random)).

I am not sure if that helps any way

MiaAltieri (commented):
Gave your changes a quick read through with some things that might be worth noting, overall it looks great! Let me know when you're ready for a thorough review :)

lib/charms/pgbouncer_operator/v0/pgb.py
```diff
@@ -82,7 +82,7 @@ def __init__(
         elif isinstance(config, dict):
             self.read_dict(config)
         elif isinstance(config, PgbConfig):
-            self.read_dict(deepcopy(config))
+            self.read_dict(config)
```
marcoppenheimer, line 85: Could this be an or?

```python
if isinstance(config, (dict, PgbConfig)):
    self.read_dict(config)
```

src/charm.py
```diff
@@ -113,15 +123,23 @@ def _on_update_status(self, _) -> None:
...
                     return
-            self.unit.status = ActiveStatus()
+
+            if self._has_backend_relation():
+                # All is well, set ActiveStatus
+                self.unit.status = ActiveStatus()
+            else:
+                # If we don't have any backend, this charm doesn't serve a purpose, and therefore
+                # should be related to one or removed.
+                self.unit.status = BlockedStatus("waiting for backend database relation")
```
marcoppenheimer, line 133: Why is this repeated? 
Is it because `BlockedStatus` wouldn't persist on a failed `_on_start`, and `_on_updated_status` would run anyway?

src/relations/backend_db_admin.py
```diff
@@ -0,0 +1,162 @@
...
+            dbs[standby_name] = pgb.parse_kv_string_to_dict(standby)
+
+        # Remove old standby information
+        for db in list(dbs.keys()):
+            if db[:21] == STANDBY_PREFIX and db not in standby_names:
+                del dbs[db]
+
+        self.charm._render_service_configs(cfg, reload_pgbouncer=True)
+
+    def _on_relation_departed(self, departed_event: RelationDepartedEvent):
```
marcoppenheimer, line 133: What are you trying to do with this event handle?
If you could explain it simply that would be really helpful :)

tests/integration/relations/test_backend_db_admin.py
```diff
@@ -0,0 +1,102 @@
...
+from tests.integration import helpers
+
+logger = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+APP_NAME = METADATA["name"]
+
+
+@pytest.mark.abort_on_fail
+async def test_create_backend_db_admin_legacy_relation_slowtest(ops_test: OpsTest):
```
marcoppenheimer, line 21: I think you could use `pytest.mark` here.

```python
@pytest.mark.slow
```

Then in your tox:


```
pytest ... -m "not slow"
```

tests/integration/relations/test_backend_db_admin.py
```diff
@@ -0,0 +1,102 @@
...
+            apps=[pg], status="active", timeout=1000, wait_for_exact_units=2
+        ),
+        ops_test.model.wait_for_idle(
+            apps=[APP_NAME], status="active", timeout=1000, wait_for_exact_units=1
+        ),
+    )
+    cfg = await helpers.get_cfg(unit)
+    # Now there are two postgres units, and the config reflects this.
+    assert list(cfg["databases"].keys()) == ["pg_master", "pgb_postgres_standby_0"]
+    assert "pgb_postgres_standby_1" not in cfg["databases"].keys()
```
marcoppenheimer, line 86: I would move this into it's own test.

Name the first block as `..._scale_up`, this as `..._scale_down`, and below `..._single_unit`, finally `..._no_units_no_master` or something. Helps separate concerns.

tests/integration/test_charm.py
```diff
@@ -34,21 +34,28 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
     )
-    # pgbouncer start command has to be successful for status to be active, and it fails if config
-    # is invalid. However, this will be tested further once health monitoring is implemented.
-    await ops_test.model.wait_for_idle(apps=[APP_NAME], status="active", timeout=1000)
+    # Pgbouncer enters a blocked status without a postgres backend database relation
+    await ops_test.model.wait_for_idle(apps=[APP_NAME], status="blocked", timeout=1000)
+    assert (
+        ops_test.model.units[f"{APP_NAME}/0"].workload_status_message
+        == "waiting for backend database relation"
+    )
```
marcoppenheimer, line 42: Maybe remove this, as it's only testing the message which is likely to change.
Maybe assert there is a message at all, rather than the specific wording.

tests/unit/relations/test_backend_db_admin.py
```diff
@@ -0,0 +1,74 @@
...
+
+        del mock_event.relation.data["mock_unit"]["standbys"]
+
+        self.relation._on_relation_changed(mock_event)
+
+        rendered_cfg = _render.call_args[0][0]
+        expected_cfg = pgb.PgbConfig(pgb.DEFAULT_CONFIG)
+        expected_cfg["databases"]["pg_master"] = pgb.parse_kv_string_to_dict(TEST_UNIT["master"])
+        self.assertEqual(expected_cfg.render(), rendered_cfg.render())
+        self.assertNotIn(f"{STANDBY_PREFIX}0", rendered_cfg.keys())
```
marcoppenheimer, line 59: Is this a separate test? I can't really tell, but if it's covering a different case, I'd split it up.
No harm in bloated test-code :), DRY is hard to maintain while also being readable.

tests/unit/relations/test_backend_db_admin.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @patch("charm.PgBouncerCharm._read_pgb_config", return_value=pgb.PgbConfig(pgb.DEFAULT_CONFIG))
+    @patch("charm.PgBouncerCharm._render_service_configs")
+    def test_on__relation_departed(self, _render, _read):
```
marcoppenheimer, line 63: Is the dunder here intentional?

marcoppenheimer (approved):
Looks fine, minor nits and a question around an event, but otherwise looks solid!

src/charm.py
WRFitch: The check is duplicated because I want to provide instant feedback about the state of the relation. The update-status hook fires every 5 minutes by default, but I want a charm admin to immediately understand that this charm won't work without the postgres charm. I can extract this into a different method, maybe the `_has_backend_relation()` function if you want?

tests/integration/relations/test_backend_db_admin.py
WRFitch: Great suggestion, but I've made a small change: I'm using `@pytest.mark` as you suggest, but with `smoke` and `legacy_relation` instead.

tests/unit/relations/test_backend_db_admin.py
WRFitch: no, removed.

src/charm.py
marcoppenheimer: Nah, I like that it's explicit, just didn't understand the repetition :)

tests/integration/test_charm.py
marcoppenheimer: If it's that important a wording, yes I think that's a sensible solution :)

src/relations/backend_db_admin.py
WRFitch: I've refactored this handle and added a `relation_broken` handler - it should be more logically laid out (and work properly) now, but if it's still unclear: 
- relation-departed should update the config to match how replicas are removed
- relation-broken removes everything postgres from the config

tests/integration/relations/test_backend_db_admin.py
WRFitch: Agreed, I also updated the scale_down test to delete the master unit to ensure that all works ok.

tests/unit/relations/test_backend_db_admin.py
WRFitch: This part of the test verifies the config correctly removes standby information that already exists in the config. I've added a comment to elucidate, but I could make a separate test with a config that already has the necessary information, then check it changes as expected if that works?

src/relations/backend_db_admin.py
```diff
@@ -0,0 +1,201 @@
...
+
+        self._update_standbys(cfg, standbys)
+
+        self.charm._render_service_configs(cfg, reload_pgbouncer=True)
+
+    def _update_standbys(self, cfg: PgbConfig, standbys: List[str]):
+        """Updates standby list to match new relation data."""
+        dbs = cfg["databases"]
+
+        standby_names = []
```
shayancanonical, line 168: where is this variable used?

src/relations/backend_db_admin.py
```diff
@@ -0,0 +1,201 @@
...
+        for idx, standby in enumerate(standbys):
+            standby_name = f"{STANDBY_PREFIX}{idx}"
+            standby_names.append(standby_name)
+            dbs[standby_name] = pgb.parse_kv_string_to_dict(standby)
```
shayancanonical, line 172: wont this overwrite the `standby_name` with every iteration of the loop?

src/relations/backend_db_admin.py
WRFitch: This is populated with each standby_name in the following loop, and then checked in the next loop when we iterate through existing in the pgbouncer config, deleting old standbys that are no longer in the relation databag. I could probably rename these variables to be more easily understood, though.

src/relations/backend_db_admin.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        logger.info("database change detected - updating config")
+        logger.warning(
+            "DEPRECATION WARNING - backend-db-admin is a legacy relation, and will be deprecated in a future release. "
+        )
+
+        postgres_data = change_event.relation.data.get(change_event.unit)
+        # if not postgres_data:
+        #     # if change_event.unit doesn't exist,
+        #     return
```
marcoppenheimer, line 118: Is this meant to be commented?

src/relations/backend_db_admin.py
marcoppenheimer: Much clearer!

tests/unit/relations/test_backend_db_admin.py
marcoppenheimer: That's clear, makes sense.

marcoppenheimer (approved):

src/relations/backend_db_admin.py
WRFitch: nope, I'll delete that now.

PR #15, opened 2022-08-19
Updated libs for k8s feature parity upgrade

WRFitch:
## Proposal
Adds a couple of new libraries that implement necessary functionality for upgrading this charm to have feature parity with the k8s charm, as well as some quality-of-life changes for the feature parity upgrade. 

## Context
- These libraries are **not new code, and don't need to be thoroughly reviewed**, but adding them in their own PR makes the next PR less huge. 
- There's been an update to the old postgres charm that's likely to be the cause of some integration test errors. Since the next PR removes this relation, I've chosen to skip all integration tests that rely on it. 

## Release Notes
- Added data_platform_libs.database_requires lib
- updated postgresql_k8s lib
- added pgbouncer_k8s.pgb lib. 
  - this behaves differently to the pgbouncer-operator lib, and updating it effectively means implementing the new version of auth, so pgbouncer-operator.pgb will be removed in the next PR. 
- Added integration test labels to break tests down into smaller runs. 
- Added ops_test.fast_forward() to integration tests, to speed them up. 
- Skipped broken tests that will be removed imminently.


.github/workflows/ci.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: lxd
-          lxd-channel: 4.24/stable # patch until https://bugs.launchpad.net/juju/+bug/1968849 is resolved
+          lxd-channel: 5.4/stable
```
marcoppenheimer, line 39: For my understanding, could you explain what this resolves?

tests/integration/helpers.py
```diff
@@ -66,7 +66,10 @@ async def get_unit_cores(unit: str) -> int:
     """
     get_cores_from_unit = await unit.run('python3 -c "import os; print(os.cpu_count())"')
     cores = get_cores_from_unit.results.get("Stdout")
-    return int(cores)
+    if cores is not None:
+        return int(cores)
+    else:
+        raise Exception(get_cores_from_unit.results)
```
marcoppenheimer, line 72: Not important enough to warrant a change, but you can also run a subprocess with:

`nproc --all` returning an int/string.

marcoppenheimer: Personally I see value in integration tests not using the same deps as the code being tests (i.e Python and `ops` code), but it's really not an important issue.

tox.ini
```diff
@@ -82,22 +82,59 @@ commands =
 description = Run integration tests
 deps =
     pytest
-    juju
+    juju==2.9.11 # juju 3.0.0 has issues with retrieving action results
```
marcoppenheimer, line 85: Good to know!

tox.ini
```diff
@@ -82,22 +82,59 @@ commands =
...
     psycopg2-binary
     -r{toxinidir}/requirements.txt
 commands =
     pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs} \
         --durations=0
 
-[testenv:integration-legacy-relations]
-description = run integration tests, skipping the slow ones.
+[testenv:dev-integration]
+description = Run integration tests marked dev
```
marcoppenheimer, line 94: Am I correct in assuming that these `tox`'s are for local development? 
If so, worth adding that to the description, minor. 

Also, the `ci.yaml` calls `tox -e integration`, which going from this means to run all non-`skip`ed tests. Is this what we want? Or were we meant to only run a specific integration `tox` during the CI?

marcoppenheimer (approved):
0 issues, mostly just questions for my sake. Good work!

tox.ini
WRFitch: Generally yes, but the plan is to use them for parallel test runs in the CI as well. For now, I haven't bothered to implement that, but I can do that in this PR if you want

.github/workflows/ci.yaml
WRFitch: Nothing in particular, I just wanted to keep it up-to-date with whatever I'm running locally, at least in terms of major versions.

tests/integration/helpers.py
WRFitch: Using python here felt like a kludge, so I'll give that a try. Thanks!

tests/integration/test_charm.py
```diff
@@ -64,6 +64,7 @@ async def test_change_config(ops_test: OpsTest):
     )
 
     # The config changes depending on the amount of cores on the unit, so get that info.
+    logging.error(unit)
```
marceloneppel: This can be changed to `logger.info` or removed.

marceloneppel (approved):
LGTM!

marceloneppel (approved):

marcoppenheimer (approved):

PR #16, opened 2022-08-30
Preliminary updates for relation PR

WRFitch:
## Proposal
Currently the relation PR is looking at a diff of 3500 lines, so I'm splitting some of the changes out into their own PRs. 

This PR completes a lot of small changes necessary to make the updated relations work.

## Context
- All these changes have already been reviewed on the k8s charm, so you don't have to go into fastidious detail on this PR unless you really really want to. 

## Release Notes
- Updated lib versions
- Downgraded file permissions
- Renamed some functions for compatibility with k8s relation libraries
- Added socket reuse to pgbouncer command, so when it restarts it'll use the same sockets without an interruption in service
- Added some new integration test helpers. These are currently unused, but will replace the existing integration test helpers. 
- Deleted unit tests for the obsolete pgb.py lib, since it's now owned by the k8s charm. 
  - The lib itself can't be deleted because it's still used by the charm. 
  - Also deleted some relevant test files

## Testing
- No code changes, so testing is unnecessary.


src/pgbouncer@.service
```diff
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 Type=simple
 User=postgres
-ExecStart=/sbin/pgbouncer /var/lib/postgresql/pgbouncer/instance_%i/pgbouncer.ini
+ExecStart=/sbin/pgbouncer -R /var/lib/postgresql/pgbouncer/instance_%i/pgbouncer.ini
```
marceloneppel, line 8: Optional: a description about the flag that is being used to make it easy to debug some issues that may arise because of that?

tests/integration/helpers/helpers.py
```diff
@@ -0,0 +1,223 @@
...
+    get_databag = await ops_test.juju(
+        "show-unit",
+        unit_name,
+        "--format=json",
+    )
+    return json.loads(get_databag[1])[unit_name]
+
+
+async def cat_file_from_unit(ops_test: OpsTest, filepath: str, unit_name: str) -> str:
+    """Gets a file from the pgbouncer container of a pgbouncer application unit."""
```
marceloneppel, line 97: You can remove the mention to container, as this is VM charm, right? The same applies to the other functions from this file.

marceloneppel (dismissed):
LGTM!

src/pgbouncer@.service
WRFitch: Sure, is a comment enough?

tests/integration/helpers/helpers.py
WRFitch: good catch!

src/pgbouncer@.service
marceloneppel: Yes.

marceloneppel (approved):

.github/workflows/ci.yaml
```diff
@@ -40,10 +42,54 @@ jobs:
...
+  #     - name: Run integration tests
+  #       # TODO set a predictable model name so it can be consumed by charm-logdump-action
+  #       # For now, setting one using "-- --model testing" causes it to get lost in the CI. This can be replicated locally.
+  #       run: tox -e legacy-relation-integration
+  #     # - name: Dump logs
+  #     #   uses: canonical/charm-logdump-action@main
+  #     #   if: failure()
+  #     #   with:
+  #     #     app: pgbouncer-operator
+  #     #     model: testing
```
shayancanonical, line 95: why add commented code? also missing new line

shayancanonical (approved):

marceloneppel (approved):

shayancanonical (approved):

PR #17, opened 2022-08-30
Add backend relation files

WRFitch:
## Proposal
This PR adds the necessary files for the updated legacy relation and the new backend relation, and starts to remove unnecessary files from the old ones. 

This PR should be merged after #16 

## Context
- I'm doing this in a bunch of small PRs so the diffs aren't enormous and they'll be easier to review. 
- *Nothing has been implemented in this PR*. I've just added some of the necessary files into the repo, with no real testing. Future PRs will include tests and implementation, and will remove the old relations. 
- Most of this has already been reviewed in terms of its design, with some small changes to make it usable for machine charms. They're likely to change after implementation and testing. 

## Release Notes
- Adds modules & supporting files for backend_database and updated db relations. 
- Deletes old tests for backend_database modules 

## Testing
No tests, since this technically hasn't been implemented.

PR #18, opened 2022-08-31
Update backend prelim integrate relations

WRFitch:
## Proposal
Ports new backend-database, db, and db-admin relations into the VM charm. This PR relies on #17 

## Context
- All these changes have already been reviewed on the k8s charm, so you don't have to go into fastidious detail on this PR unless you really really want to. 
- A lot of the negative diff is deleting old tests

## Release Notes
- Add new backend-database relation to new postgres charm, replacing old backend-db-admin relation
- Add legacy db and db-admin client relations, replacing old implementations
- removed pgbouncer-operator pgb lib from this charm, now it has been replaced with pgbouncer-k8s pgb lib. 
- Removed unnecessary tests
- Reinstated relevant test runners
- Added the ability to change the port on which pgbouncer listens. 
- Renamed peer relation stub to match k8s charm. Implementation will come in a future PR. 

## Testing
- Unit tests have been updated
- Integration tests have been updated to work with new implementation. They'll be updated further in a later PR.


src/relations/db.py
```diff
@@ -269,12 +467,18 @@ def _comma_split(string) -> Iterable[str]:
...
-            unit
-            for unit in relation.data
-            if isinstance(unit, Unit) and not unit.name.startswith(self.model.app.name)
-        ]
+        return ",".join(
+            sorted(
+                [
+                    unit.name
+                    for unit in relation.data
+                    if isinstance(unit, Unit) and not unit.app != self.charm.app
```
marceloneppel, line 475: This (the double negative) I know that will be addressed in the next PR.

marceloneppel (commented):
I left some comments about the implementation.

tests/integration/relations/test_backend_database.py
```diff
@@ -54,8 +52,8 @@ async def test_relate_pgbouncer_to_postgres(ops_test: OpsTest):
         wait_for_relation_joined_between(ops_test, PG, PGB)
         await ops_test.model.wait_for_idle(apps=[PGB, PG], status="active", timeout=1000),
 
-        cfg = await get_cfg(ops_test)
-        logging.info(cfg.render())
+        cfg = await get_cfg(ops_test, "pgbouncer-operator/0")
```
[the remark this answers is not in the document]
WRFitch: In this instance, we're fine to use the default ini, stored in the pgbouncer home directory.

marceloneppel (dismissed):

src/charm.py
```diff
@@ -151,9 +134,40 @@ def _on_config_changed(self, _) -> None:
             self._cores,
         )
 
-        cfg["pgbouncer"]["listen_addr"] = str(self.unit_ip)
```
MiaAltieri, line 154: Question just for my understanding: Why isn't this needed anymore?

src/charm.py
```diff
@@ -358,35 +279,65 @@ def _install_apt_packages(self, packages: List[str]):
...
         os.chown(path, uid=u.pw_uid, gid=u.pw_gid)
 
-    def _has_backend_relation(self) -> bool:
-        """Returns whether or not this charm is related to a postgresql backend.
-
-        TODO this will be updated to include the new backend relation once it is written.
-        """
-        legacy_backend_relation = self.model.get_relation(BACKEND_DB_ADMIN)
-        return legacy_backend_relation is not None
+    def delete_file(self, path: str):
```
MiaAltieri, line 299: might be good in a helpers file

src/relations/db.py
```diff
@@ -3,58 +3,97 @@
...
-│           │          port │ 6432                                              │        │
-│           │      standbys │ dbname=cli_standby host=10.101.233.10             │        │
-│           │               │ password=jnT4LxNPPrssscxGYmGPy4FKjRNXCn4NL2Y32jqs │        │
-│           │               │ port=6432 user=db_85_psql                         │        │
-│           │         state │ master                                            │        │
-│           │          user │ db_85_psql                                        │        │
-│           │       version │ 12                                                │        │
-└───────────┴───────────────┴───────────────────────────────────────────────────┴────────┘
-
+TODO update relation data to match VM IPs, not k8s hostnames
```
MiaAltieri, line 10: maybe you can do this before you merge

src/relations/db.py
```diff
@@ -88,113 +130,243 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
 
-        Takes information from the db relation databag and copies it into the pgbouncer.ini
-        config.
+        If the backend relation is fully initialised and available, we generate the proposed
+        database and create a user on the postgres charm, and add preliminary data to the databag.
         """
-        if not self.charm.unit.is_leader():
+        if not self.charm.backend.postgres:
+            # We can't relate an app to the backend database without a backend postgres relation
+            wait_str = "waiting for backend-database relation to connect"
```
MiaAltieri, line 141: nit: this string is used in multiple functions, might be best to make it a constant

src/relations/db.py
```diff
@@ -88,113 +130,243 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
+        if not database:
+            # If the database field isn't populated, the remote app may be using the unit databag.
+            remote_databag = join_event.relation.data[join_event.unit]
+            database = remote_databag.get("database")
+
+            if not database:
+                # If there's nothing in either databag, break.
+                logger.warning("No database name provided in app databag")
+                join_event.defer()
+                return
```
MiaAltieri, line 173: Minor simiplication to help with indention complexity.

Followup questions: 

1. Why might it be provided in either app or unit data? Is that something we can control, because it would be nice to have consistency.
2. Should we go into WaitingStatus or is that overkill?

```suggestion
        remote_app_databag = join_event.relation.data[join_event.app]
        remote_unit_databag = join_event.relation.data[join_event.unit]

        database = remote_app_databag.get("database")
        # If the database field isn't populated, the remote app may be using the unit databag.
        database = remote_unit_databag.get("database") if not databag
        if not database:
             # If there's nothing in either databag, break.
             logger.warning("No database name provided in app or unit databag")
             join_event.defer()
             return
```

src/relations/db.py
```diff
@@ -88,113 +130,243 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
+
+        # Do not allow apps requesting extensions to be installed.
```
MiaAltieri, line 175:
```suggestion
        # Do not allow apps requesting extensions to be installed. Extensions cannot be requested through relations, instead they should be installed through a database charm config.
```

src/relations/db.py
```diff
@@ -88,113 +130,243 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
+        # been created, so we defer this event until the databag is populated.
+        pgb_app_databag = change_event.relation.data[self.charm.app]
+        database = pgb_app_databag.get("database")
+        user = pgb_app_databag.get("user")
+        password = pgb_app_databag.get("password")
+
+        if None in [database, user, password]:
+            logger.warning(
+                "relation not fully initialised - deferring until join_event is complete"
+            )
```
MiaAltieri, line 256: do you want a waiting status

src/relations/db.py
```diff
@@ -88,113 +130,243 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
-        for databag in [pgb_app_databag, pgb_unit_databag]:
-            updates = {
+        self.update_port(change_event.relation, self.charm.config["listen_port"])
+        self.update_postgres_endpoints(change_event.relation, reload_pgbouncer=True)
+        self.update_databag(
+            change_event.relation,
+            {
                 "allowed-subnets": self.get_allowed_subnets(change_event.relation),
                 "allowed-units": self.get_allowed_units(change_event.relation),
-                "host": f"http://{master_host}",
```
MiaAltieri, line 149: Question for my understanding: why is this removed?

src/relations/db.py
```diff
@@ -88,113 +130,243 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
+            relation,
+            {
+                "master": dbconnstr,
+                "port": str(port),
+                "standbys": dbconnstr,
+            },
+        )
+
+    def update_postgres_endpoints(self, relation: Relation, reload_pgbouncer: bool = False):
+        """Updates postgres replicas."""
```
MiaAltieri, line 307: change

src/relations/db.py
```diff
@@ -88,113 +130,243 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
+
+        for databag in databags:
+            databag.update(updates)
 
-            standbys.append(pgb.parse_dict_to_kv_string(standby))
+    def _generate_username(self, event):
+        """Generates a unique username for this relation."""
+        app_name = self.charm.app.name
+        relation_id = event.relation.id
+        model_name = self.model.name
```
MiaAltieri, line 355: clever

src/relations/db.py
```diff
@@ -232,22 +400,53 @@ def _on_relation_departed(self, departed_event: RelationDepartedEvent):
...
+            cfg["databases"].pop(database, None)
+            cfg["databases"].pop(f"{database}_standby", None)
+
+        cfg.remove_user(user)
+        self.charm.render_pgb_config(cfg, reload_pgbouncer=True)
+        try:
+            self.charm.backend.postgres.delete_user(user)
+        except PostgreSQLDeleteUserError:
+            # We've likely lost connection at this point, and can't do anything about a trailing
+            # user.
```
MiaAltieri, line 448: I think something to this effect should be logged

MiaAltieri (commented):
Reviewed charm.py and db.py as requested. Looks good, minor comments

src/charm.py
WRFitch: listen_addr is now a wildcard, because tracking them all individually isn't necessary and makes scaling harder.

src/charm.py
WRFitch: Good shout, I'm planning some quality-of-life refactors once all the features for this sprint are done so I might leave this until then?

src/relations/db.py
WRFitch: 1. One of the databags (I believe it was app) was added after the unit databag, so different charms handle it differently depending on when they were written, and there's no guarantee which one is used because the database name is provided by the client charm. After this point, we spam our own databags with all the info we need, because client charms can't be trusted. 
2. Waitingstatus should be fine, I've added it in. 

Good suggestion though, I've tightened it up further using some walrus operators.

src/relations/db.py
WRFitch: This line specifically? It's just been moved down to L268

shayancanonical (dismissed):

shayancanonical (approved):

marceloneppel (approved):

PR #20, opened 2022-09-13
Scaling

WRFitch:
## Proposal
Adds scaling to VM charm

Jira issue: [DPE-471](https://warthogs.atlassian.net/browse/DPE-471)

## Context
- A lot of this code is ported from the k8s charm, so it's less important for it to be reviewed unless you really like reviewing things. 
  - relations/peers.py
  - pgb.py

## Release Notes
- Adds scaling to pgbouncer vm charm
- adds dedicated integration tests & CI
  - I've also reduced the timeout on failing integration test status checks, and I may reduce it further. Generally, most of the tests have failed if it takes longer than a few minutes to connect pg to pgb, so waiting 10/15 minutes is unnecessary. 
- updated relation docstring examples
- Added a better check for pgbouncer status

## Testing
- Added new integration tests that test scaling directly
- Added new integration tests to db relation tests##
- added unit tests


src/relations/backend_database.py
```diff
@@ -161,12 +168,16 @@ def _on_relation_departed(self, event: RelationDepartedEvent):
...
         """Handle backend-database-relation-broken event.
 
         Removes all traces of this relation from pgbouncer config.
         """
-        cfg = self.charm.read_pgb_config()
+        try:
+            cfg = self.charm.read_pgb_config()
+        except FileNotFoundError:
+            event.defer()
+            return
```
marceloneppel, line 180: Can the defer call be removed? (I am thinking whether the file will exist earlier, to create the user and the database, and then later it wouldn't exist anymore).

tests/integration/relations/test_peers.py
```diff
@@ -0,0 +1,133 @@
...
+        await ops_test.model.wait_for_idle(
+            apps=[PGB], status="blocked", timeout=600, wait_for_exact_units=3
+        ),
+
+
+@pytest.mark.scaling
+@pytest.mark.abort_on_fail
+@pytest.mark.run(order=2)
+async def test_scaled_relations(ops_test: OpsTest):
+    """Test that the pgbouncer and postgres charms can relate to one another."""
```
marceloneppel, line 46: This description can probably be changed to include the client app, because the relation between it and pgbouncer is also tested.

tests/integration/relations/test_peers.py
```diff
@@ -0,0 +1,133 @@
...
+            ops_test.model.wait_for_idle(
+                apps=[PG], status="active", timeout=600, wait_for_exact_units=3
+            ),
+        )
+
+
+@pytest.mark.scaling
+@pytest.mark.run(order=3)
+async def test_scaling(ops_test: OpsTest):
+    """Test data is replicated to new units after a scale up."""
```
marceloneppel, line 91: Is it worth to check PgBouncer config to ensure the data is replicated?

src/charm.py
```diff
@@ -142,22 +140,38 @@ def _on_config_changed(self, _) -> None:
...
+            logger.warning(wait_str)
+            self.unit.status = WaitingStatus(wait_str)
+            return False
+
+        if not self.backend.postgres:
+            # We can't relate an app to the backend database without a backend postgres relation
+            backend_wait_msg = "waiting for backend-database relation to connect"
+            logger.warning(backend_wait_msg)
+            self.unit.status = BlockedStatus(backend_wait_msg)
+            return False
```
marceloneppel, line 158: Optional: maybe move the calls related to setting the unit status to the event handlers, if possible. But if it complicates a lot, ignore it.

marceloneppel (commented):
Amazing work, Will. I liked your changes on timeouts, they seem to be a good addition to the tests.

shayancanonical (dismissed):
lgtm, pending marcelo's comments

src/charm.py
WRFitch: It seems cleanest to change this method's return value to a `Status` of some kind, then do something like this in the event handlers: 
```python
self.charm.unit.status = self.charm.get_status()
```
What do you think?

tests/integration/relations/test_peers.py
WRFitch: This test primarily exists to validate that we can scale in whatever direction we like without damaging anything, but it wouldn't hurt to ensure all the units have a valid config

src/charm.py
marceloneppel: Yeah, it seems ok.

src/relations/backend_database.py
marceloneppel: thanks for the explanation, Will!

marceloneppel (dismissed):
Pretty solid!

src/relations/peers.py
```diff
@@ -110,15 +112,12 @@ def unit_databag(self):
...
+        """Fetch current set of peers IPs.
 
         Returns:
             A set of peers addresses (strings).
         """
-        # Get all members IPs and remove the current unit IP from the list.
-        addresses = {self._get_unit_ip(unit) for unit in self.relation.units}
-        addresses.add(self.charm.unit_ip)
-        return addresses
+        return {self._get_unit_ip(unit) for unit in self.relation.units}
```
marceloneppel: Shouldn't you add something similar to what was added on https://github.com/canonical/pgbouncer-k8s-operator/blob/a5d32d72ef3f447878ad575255477cb5b3465b6e/src/relations/peers.py#L125?

marceloneppel (dismissed):
Looks awesome! I have just one question (after reviewing https://github.com/canonical/pgbouncer-k8s-operator/pull/18) and thinking about the issue we had with the IPs earlier today.

src/relations/peers.py
WRFitch: This has been added - thanks for the suggestion.

marceloneppel (dismissed):

src/relations/peers.py
```diff
@@ -0,0 +1,302 @@
...
+            key: key to access data
+        """
+        if scope == "unit":
+            self.unit_databag.pop(key, None)
+            return
+        elif scope == "app":
+            self.app_databag.pop(key, None)
+            return
+        else:
+            raise RuntimeError("Unknown secret scope.")
```
zmraul, line 246: You can also just use `set_secret` with an empty string as value.

zmraul (approved):
Looks neat :)

marceloneppel (approved):

zmraul (approved):

PR #24, opened 2022-10-04
updated readonly endpoints

WRFitch:
## Proposal
When pointing to readonly endpoints, pgbouncer would just pick a random standby node and point to that. This PR updates the standby node in the pgb config to point to all postgres standby nodes, switching between them in a round robin. 

## Context
[docs](https://www.pgbouncer.org/config.html#host)

## Testing
- Unit tests have been updated, all integration tests currently pass


welpaolo (approved):
LGTM

marceloneppel (approved):
LGTM!

PR #26, opened 2022-10-05
Revert "updated readonly endpoints"

WRFitch:
## Proposal
Reverts canonical/pgbouncer-operator#24

## Context
The feature that this PR relies on is only implemented in pgbouncer version 1.17. This'll be re-implemented once the pgbouncer snap is available

### DIFF
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -84,7 +84,7 @@ Some example relation data is below. All values are examples, generated in a run
 └──────────────────┴─────────────────────┴─────────────────────────────────────────────────────────────────┘
 """  # noqa: W505
 import logging
-from typing import Dict, Iterable, List
+from typing import Dict, Iterable
 
 from charms.pgbouncer_k8s.v0 import pgb
 from charms.postgresql_k8s.v0.postgresql import (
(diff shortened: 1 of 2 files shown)


marceloneppel (approved):

src/relations/db.py
```diff
@@ -494,16 +493,16 @@ def _generate_username(self, event):
...
 
         Though multiple readonly endpoints can be provided by the new backend relation, only one
         can be consumed by this legacy relation.
         """
         read_only_endpoints = self.charm.backend.postgres_databag.get("read-only-endpoints")
         if read_only_endpoints is None or len(read_only_endpoints) == 0:
-            return []
-        return read_only_endpoints.split(",")
+            return None
+        return read_only_endpoints.split(",")[0]
```
marcoppenheimer, line 505: Could do simply:

```python
read_only_endpoints = self.charm.backend.postgres_databag.get("read-only-endpoints", "")
return read_only_endpoints.split(",")[0] or None
```

marcoppenheimer (approved):

src/relations/db.py
WRFitch: Nice, I just want to revert this PR but I'll add this in future. Thanks for the suggestion!

PR #31, opened 2022-10-28
Add jammy support

WRFitch:
## Proposal
Builds & deploys on Jammy

## Context
- Tests have been moved to Jammy, with no provision for Focal. This is because I'm planning to deprecate focal once Jammy is implemented in the postgresql-bundle repo.

### DIFF
--- a/tests/integration/relations/test_peers.py
+++ b/tests/integration/relations/test_peers.py
@@ -33,7 +33,7 @@ async def test_deploy_at_scale(ops_test):
     # Build, deploy, and relate charms.
     charm = await ops_test.build_charm(".")
     async with ops_test.fast_forward():
-        await ops_test.model.deploy(charm, application_name=PGB, num_units=3)
+        await ops_test.model.deploy(charm, application_name=PGB, num_units=3, series="jammy")
         await ops_test.model.wait_for_idle(
             apps=[PGB], status="blocked", timeout=600, wait_for_exact_units=3
         ),
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,9 +26,6 @@ jobs:
         run: tox -e unit
 
   integration-standalone:
-    strategy:
-      matrix:
-        series: [focal, jammy]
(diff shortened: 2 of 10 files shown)


marceloneppel (dismissed):

tests/integration/relations/test_peers.py
```diff
@@ -33,7 +33,7 @@ async def test_deploy_at_scale(ops_test):
     # Build, deploy, and relate charms.
     charm = await ops_test.build_charm(".")
     async with ops_test.fast_forward():
-        await ops_test.model.deploy(charm, application_name=PGB, num_units=3)
+        await ops_test.model.deploy(charm, application_name=PGB, num_units=3, series="jammy")
```
marceloneppel, line 36: Is it worth to have something like the [parametrised series](https://github.com/canonical/postgresql-operator/blob/01936670557ba4a65473cc45f1078ae71cbb5cdf/tests/integration/test_charm.py#L32) we had some months ago on the PostgreSQL charm?

marceloneppel (approved):

tests/integration/relations/test_peers.py
WRFitch: I had actually planned to remove focal support entirely after this PR was uploaded to charmhub (and therefore usable in the bundle), making multiple series testing unnecessary, but I suppose that's a product decision

zmraul (approved):

dragomirp (approved):

delgod (approved):

taurus-forever (approved):
I am not sure why we need to remove focal, as anyway we need to learn how to support several series in parallel. Anyway, LGTM. Tnx!

WRFitch:
> I am not sure why we need to remove focal, as anyway we need to learn how to support several series in parallel. 

The idea behind supporting Jammy is to allow us to use some newer features of PGB without having to build a snap yet. The reason why I don't want to support both is because I'd still have to figure out a workaround on the focal charm, and it would also slow development since test runs would be doubled across series.

PR #32, opened 2022-11-02
Add client relation libs, and other busywork

WRFitch:
## Proposal
Preliminary updates for the new client relation.

## Release Notes
- Updated postgres lib, with minor compatibility updates
- added provider lib
- Version updates

PR #33, opened 2022-11-02
Add client relation

WRFitch:
## Proposal
Adds new client relation to VM charm. This PR depends on #32.

## Release Notes
- Adds untested new client relation.

## Testing
- Added unit tests
- Integration tests coming in a future PR


src/relations/db.py
```diff
@@ -332,7 +332,7 @@ def update_connection_info(self, relation: Relation, port: str):
             "host": self.charm.unit_ip,
         }
 
-        standby_ips = self.charm.peers.units_ips - {self.charm.peers.leader_ip}
+        standby_ips = self.charm.peers.units_ips - {self.charm.leader_ip}
```
paulomach, line 335: Is `self.charm.peers.units_ips` a set? If not, just `set(self.charm.peers.units_ips)`

src/relations/pgbouncer_provider.py
```diff
@@ -0,0 +1,266 @@
...
+            PostgreSQLCreateUserError,
+            PostgreSQLGetPostgreSQLVersionError,
+        ) as e:
+            logger.exception(e)
+            self.charm.unit.status = BlockedStatus(
+                f"Failed to initialize {self.relation_name} relation"
+            )
+            return
+
+        self.charm.peers.add_user(user, password)
```
paulomach, line 125: Why duplicating credentials when they are set on relation databag?

src/relations/pgbouncer_provider.py
```diff
@@ -0,0 +1,266 @@
...
+        self.update_postgres_endpoints(event.relation, cfg=cfg)
+        self.charm.render_pgb_config(cfg, reload_pgbouncer=True)
+
+        # Share the credentials and updated connection info with the client application.
+        self.database_provides.set_credentials(rel_id, user, password)
+        self.update_connection_info(event.relation)
+
+    def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
+        """Remove the user created for this relation, and revoke connection permissions."""
+        if not self._check_backend():
```
paulomach, line 139: Be aware that you could be receiving a `RelationBrokenEvent` from a related unit being torn down due to scale down.

paulomach (commented):
Few comments and focus on the relation broken thing!

src/relations/pgbouncer_provider.py
WRFitch: Good catch, removed.

src/relations/db.py
WRFitch: It is indeed a set :)

src/relations/db.py
```diff
@@ -417,9 +427,17 @@ def _on_relation_broken(self, broken_event: RelationBrokenEvent):
...
             self.charm.update_client_connection_info()
             return
+        if not self.charm.peers.app_databag.get(
+            f"{self.relation_name}-{broken_event.relation.id}-relation-breaking", None
+        ):
+            # This relation isn't being removed, so we don't need to do the relation teardown
+            # steps.
+            self.update_connection_info(broken_event.relation)
+            self.update_postgres_endpoints(broken_event.relation, reload_pgbouncer=True)
+            return
```
marceloneppel, line 440: Nice! Would it be good to add the [bug number or link](https://bugs.launchpad.net/juju/+bug/1979811). Maybe it's easier to find this later when the bug is solved. Or we can create a Jira issue to update this later (if the bug is resolved at some point).

src/relations/pgbouncer_provider.py
WRFitch: After checking the [docs](https://juju.is/docs/sdk/relation-events), I found that when we only receive relation-departed events where `event.app == self.charm.app` when the relation is being removed, so I added a relation-departed hook that checks this and sets a unique flag in the peer databag. If this flag is present in the relation-broken hook, then we teardown the whole relation.

src/relations/db.py
WRFitch: This is actually [documented behaviour](https://juju.is/docs/sdk/relation-name-relation-broken-event) - it's not super intuitive, though.

src/relations/pgbouncer_provider.py
```diff
@@ -0,0 +1,289 @@
...
+            return
+        if not self.charm.unit.is_leader():
+            return
+        if not self.charm.peers.app_databag.get(
+            f"{self.relation_name}-{event.relation.id}-relation-breaking", None
+        ):
+            # This relation isn't being removed, so we don't need to do the relation teardown
+            # steps.
+            self.update_connection_info(event.relation)
+            self.update_postgres_endpoints(event.relation, reload_pgbouncer=True)
```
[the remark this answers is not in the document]
WRFitch: nope, this was unnecessary :)

src/relations/db.py
```diff
@@ -407,6 +410,13 @@ def _on_relation_departed(self, departed_event: RelationDepartedEvent):
             {"allowed-units": self.get_allowed_units(departed_event.relation)},
         )
 
+        # If a departed event is dispatched to itself, this relation isn't being scaled down -
+        # it's being removed
+        if departed_event.app == self.charm.app:
+            self.charm.peers.app_databag[
```
marceloneppel, line 416: Here I think we need a check for leadership, because you are setting data in the app databag.

marceloneppel (approved):
LGTM!

src/relations/db.py
WRFitch: Added :)

paulomach (approved):
Fantastic

PR #41, opened 2022-12-12
systemd socket activation

WRFitch:
## Proposal
Adds [systemd socket activation](https://www.2ndquadrant.com/en/blog/running-multiple-pgbouncer-instances-with-systemd/)

## Context
Previously, we were having to use some workarounds to reuse ports across pgbouncer instances, where we were duplicating the config for each systemd instance of pgbouncer. By using systemd sockets and pgbouncer>=1.14, we no longer have to do this.

## Testing
Integration tests are passing, no new testing added.

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -153,9 +153,10 @@ class TestCharm(unittest.TestCase):
 
     @patch("charm.PgBouncerCharm.read_pgb_config", return_value=pgb.PgbConfig(DEFAULT_CFG))
     @patch("charm.PgBouncerCharm.render_pgb_config")
+    @patch("charm.PgBouncerCharm.update_systemd_socket")
     @patch("relations.peers.Peers.app_databag", new_callable=PropertyMock)
     @patch_network_get(private_address="1.1.1.1")
-    def test_on_config_changed(self, _app_databag, _render, _read):
+    def test_on_config_changed(self, _app_databag, _update_socket, _render, _read):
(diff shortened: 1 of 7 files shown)


tests/unit/test_charm.py
```diff
@@ -233,29 +231,15 @@ def test_read_pgb_config(self):
...
+        _render.assert_any_call(INI_PATH, default_cfg.render(), 0o700)
         _reload.assert_not_called()
         # MaintenanceStatus will exit once pgbouncer reloads.
         self.assertIsInstance(self.harness.model.unit.status, MaintenanceStatus)
 
-        self.charm.render_pgb_config(cfg, reload_pgbouncer=True)
+        self.charm.render_pgb_config(default_cfg, reload_pgbouncer=True)
+        _render.assert_any_call(INI_PATH, default_cfg.render(), 0o700)
         _reload.assert_called_once()
+        self.assertIsInstance(self.harness.model.unit.status, MaintenanceStatus)
```
marceloneppel, line 245: Shouldn't the status change according to L239's comment?

marceloneppel (approved):

taurus-forever (approved):
LGTM, please check/reply Marcelo comment.

WRFitch:
Spoken to Mykola about this after asking for some help, and this PR isn't strictly necessary: adding socket activation increases complexity for no real benefit, so we're skipping it. I did learn about some other fixes and efficiencies that I'm going to implement in another PR, however.

PR #42, opened 2022-12-16
Update systemd handling

WRFitch:
## Proposal
Some minor updates to how we handle systemd

## Release Notes
- Add a systemd target file to handle the multiple pgbouncer services we want
- updated service test to check number of pgb services in a more elegant way


WRFitch:
Closed PR because systemd targets are overcomplicating a working solution.

PR #47, opened 2023-02-16
Cache charmcraft pack container, skip unstable tests except on schedule

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-k8s-operator/pull/146

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -10,7 +10,7 @@ import pytest
 from pytest_operator.plugin import OpsTest
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
 def ops_test(ops_test: OpsTest) -> OpsTest:
     if os.environ.get("CI") == "true":
         # Running in GitHub Actions; skip build step


dragomirp (approved):

pyproject.toml
```diff
@@ -12,15 +12,7 @@ show_missing = true
...
-markers = [
-    "smoke: quick integration tests that can be run to quickly verify that the charm builds and deploys ok.",
-    "dev: a quick development marker to run one test",
-    "standalone: integration tests that relate to standalone charm function",
-    "backend: integration tests that test backend relation to postgresql",
-    "legacy_relation: integration tests that test legacy client relations",
-    "client_relation: integration tests that test modern client relations",
-    "scaling: integration tests for scaling pgbouncer",
-]
+markers = ["unstable"]
```
taurus-forever, line 15: Up 2 Marcelo here, PostgreSQL team might use some old marks... but most probably we are calling test through tox anyway.

taurus-forever (approved):
LGTM, but old markers removal is up to Marcelo.

pyproject.toml
marceloneppel: We can safely remove those marks. They are not needed anymore, including the `smoke` mark (used while developing client relations in the past).

marceloneppel (approved):

PR #56, opened 2023-03-03
Revive pgb

dragomirp:
## Proposal
Seems like the update to jammy for postgresql switch the default password hashing. This forces md5 mechanism for the pgb auth user.

### DIFF
--- a/src/relations/backend_database.py
+++ b/src/relations/backend_database.py
@@ -125,12 +125,12 @@ class BackendDatabaseRequires(Object):
             return
 
         plaintext_password = pgb.generate_password()
+        hashed_password = pgb.get_hashed_password(self.auth_user, plaintext_password)
         # create authentication user on postgres database, so we can authenticate other users
         # later on
-        self.postgres.create_user(self.auth_user, plaintext_password, admin=True)
+        self.postgres.create_user(self.auth_user, hashed_password, admin=True)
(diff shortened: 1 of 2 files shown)


src/relations/backend_database.py
```diff
@@ -125,12 +125,12 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
             return
 
         plaintext_password = pgb.generate_password()
+        hashed_password = pgb.get_hashed_password(self.auth_user, plaintext_password)
         # create authentication user on postgres database, so we can authenticate other users
         # later on
-        self.postgres.create_user(self.auth_user, plaintext_password, admin=True)
+        self.postgres.create_user(self.auth_user, hashed_password, admin=True)
```
dragomirp, line 131: We should check and change this for pgb-k8s so it doesn't break when we switch to pg 14 in the OCI.

marceloneppel: Cool! On the k8s charm, the new auth mechanism was reverted on https://github.com/canonical/postgresql-k8s-operator/pull/52/files, so it's working because of that. I'll ask about scram on this week database office hours sessions, because PgBouncer default auth type is still md5 in the charm.

marceloneppel (approved):

taurus-forever (approved):
Hm... let's discuss it in details on the sync today.
We can merge this to fix CI/CD.

PR #58, opened 2023-03-24
[MISC] Add database field in the provider-side databag

dragomirp:
## Proposal
Adding the database field in the provider side. This is to follow the interface signature change https://github.com/canonical/charm-relation-interfaces/pull/43


taurus-forever (approved):
Tnx for the fast fix!

PR #73, opened 2023-05-15
[DPE-1648] Switch to charmed snap

dragomirp:
# Issue
Change the charm to use the charmed-postgresql snap

# Solution

# Context
* A parallel PR https://github.com/canonical/charmed-postgresql-snap/pull/10 for the snap
* Services are still created by the charm

# Testing

# Release Notes
Switch charmed postgresql snap


lib/charms/operator_libs_linux/v0/passwd.py
dragomirp, line 1: Doesn't seem to be used for anything

src/charm.py
```diff
@@ -72,43 +80,60 @@ def _on_install(self, _) -> None:
...
+            selected_snap = cache["charmed-postgresql"]
+            selected_snap.stop(services=["pgbackrest-service"], disable=True)
+        except snap.SnapError as e:
+            error_message = "Failed to start patroni snap service"
+            logger.exception(error_message, exc_info=e)
+
+        pg_user = pwd.getpwnam(PG_USER)
+        app_conf_dir = f"{PGB_CONF_DIR}/{self.app.name}"
+        app_log_dir = f"{PGB_LOG_DIR}/{self.app.name}"
+        app_temp_dir = f"/tmp/snap-private-tmp/snap.charmed-postgresql/tmp/{self.app.name}"
```
dragomirp, line 102: Temp files would be wiped on restart, so we should be recreating them on start. I will address that when doing COS, because having snap services running changes the behaviour.

templates/pgbouncer.service.j2
```diff
@@ -4,12 +4,11 @@ After=network.target
 
 [Service]
 Type=simple
-User=postgres
```
dragomirp, line 7: Cannot run setpriv as a regular user.

templates/pgbouncer.service.j2
```diff
@@ -4,12 +4,11 @@ After=network.target
...
 # -R flag lets separate pgbouncer instances reuse sockets from previous instances, so on restart
 # they'll reuse the same sockets, preserving connections.
-ExecStart=/sbin/pgbouncer -R /var/lib/postgresql/pgbouncer/{{ app_name }}/instance_%i/pgbouncer.ini
-ExecReload=/bin/kill -HUP $MAINPID
-KillSignal=SIGINT
+ExecStart=/snap/bin/charmed-postgresql.pgbouncer-server -R {{ conf_dir }}/{{ app_name }}/instance_%i/pgbouncer.ini
+ExecStop=pkill -INT -F /tmp/snap-private-tmp/snap.charmed-postgresql/tmp/{{ app_name }}/instance_%i/pgbouncer.pid
+ExecReload=pkill -HUP -F /tmp/snap-private-tmp/snap.charmed-postgresql/tmp/{{ app_name }}/instance_%i/pgbouncer.pid
```
dragomirp, line 11: This is hacky, but the setpriv script isn't passing SIGINT to pgbouncer. Any suggestions how to do this better?

templates/pgbouncer.service.j2
```diff
@@ -4,12 +4,11 @@ After=network.target
 
 [Service]
 Type=simple
-User=postgres
 # -R flag lets separate pgbouncer instances reuse sockets from previous instances, so on restart
 # they'll reuse the same sockets, preserving connections.
```
delgod, line 8: can we please avoid the creation of tmp and logs directories in charm code but create them here, like this
```suggestion
# they'll reuse the same sockets, preserving connections.
ExecStartPre=-/usr/bin/install -o snap_daemon -g snap_daemon -m 700 -d \
        /var/snap/charmed-postgresql/common/var/log/pgbouncer/{{ app_name }}/ \
        /tmp/snap-private-tmp/snap.charmed-postgresql/tmp/{{ app_name }}/
```

src/charm.py
```diff
@@ -72,43 +80,60 @@ def _on_install(self, _) -> None:
...
+        os.mkdir(app_conf_dir, 0o700)
+        os.chown(app_conf_dir, pg_user.pw_uid, pg_user.pw_gid)
+        os.mkdir(app_log_dir, 0o700)
```
delgod, line 105: 1) Don't create intermediate paths. Use `os.makedirs(name, mode=0o700, exist_ok=True)`
2) Create config path, near code which creates config file

delgod (changes requested):

templates/pgbouncer.service.j2
dragomirp: I'll give it a try, but `/tmp/snap-private-tmp/snap.charmed-postgresql/tmp` is itself created only when the snap is running and the snap will complain if permissions are not as expected. I'll most probably need to tweak to additionally those.

templates/pgbouncer.service.j2
marceloneppel: Does setpriv's `--pdeathsig` help with this? I haven't tested.

marceloneppel (approved):

delgod (approved):

PR #74, opened 2023-05-22
[DPE-1778] COS

dragomirp:
## Issue
Add COS integration for Pgbouncer

## Solution
* Create a service for the snapped exporter, so that multiple ppgbouncers can coexist within the same principle
* Add a graphana dashboard definition
* Add relation for the grafana-agent charm


config.yaml
```diff
@@ -10,6 +10,12 @@ options:
       listen_port in pgbouncer config.
     type: int
 
+  metrics_port:
+    default: 9127
+    description: |
+      The port on which the pgbouncer prometheus exporter serves metrics.
+    type: int
+
```
dragomirp, line 18: We need to be able to serve metrics on different ports to be able to have multiple subordinate pgbs within a single principal.

src/charm.py
```diff
@@ -156,6 +186,9 @@ def _on_start(self, _) -> None:
 
         Runs pgbouncer through systemd (configured in src/pgbouncer.service)
         """
+        # Done first to instantiate the snap's private tmp
+        self.unit.set_workload_version(self.version)
+
```
dragomirp, line 191: Snap refuses to run if we create it's private `tmp` dir manually. Setting the version first will create it, so instance socket dirs can be recreated.

src/relations/backend_database.py
```diff
@@ -131,15 +138,26 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
...
 
         cfg = self.charm.read_pgb_config()
-        cfg.add_user(user=event.username, admin=True)
+        cfg.add_user(user=self.stats_user, stats=True)
         cfg["pgbouncer"][
             "auth_query"
         ] = f"SELECT username, password FROM {self.auth_user}.get_auth($1)"
         cfg["pgbouncer"]["auth_file"] = f"{PGB_CONF_DIR}/{self.charm.app.name}/{AUTH_FILE_NAME}"
         self.charm.render_pgb_config(cfg)
+        self.charm.render_prometheus_service()
```
dragomirp, line 160: Exporter will fail if there are issues with pgb services, so it's initialised with the backend relation.

src/relations/backend_database.py
```diff
@@ -217,14 +235,16 @@ def _on_relation_broken(self, event: RelationBrokenEvent):
 
         if self.postgres:
             cfg.remove_user(self.postgres.user)
-        cfg["pgbouncer"].pop("auth_user", None)
```
dragomirp, line 220: I don't think this is used for anything and it's not present in the user list.

src/relations/backend_database.py
```diff
@@ -131,15 +138,26 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
...
+        if not (
+            monitoring_password := self.charm.peers.get_secret("app", MONITORING_PASSWORD_KEY)
+        ):
+            monitoring_password = pgb.generate_password()
+            self.charm.peers.set_secret("app", MONITORING_PASSWORD_KEY, monitoring_password)
+        hashed_monitoring_password = pgb.get_hashed_password(self.stats_user, monitoring_password)
+
+        self.charm.render_auth_file(
+            f'"{self.auth_user}" "{hashed_password}"\n"{self.stats_user}" "{hashed_monitoring_password}"'
+        )
```
dragomirp, line 151: Keep the same user and password. This should work with different backends in principle.

marceloneppel (approved):
LGTM

marceloneppel:
Maybe Paulo's feedback is worth in the parts really specific to COS.

src/charm.py
```diff
@@ -226,8 +266,14 @@ def check_status(self) -> Union[ActiveStatus, BlockedStatus, WaitingStatus]:
             logger.warning(backend_wait_msg)
             return BlockedStatus(backend_wait_msg)
 
+        prom_service = f"{PGB}-{self.app.name}-prometheus"
+        services = [*self.pgb_services]
+
+        if self.backend.postgres:
```
paulomach, line 272: @dragomirp does the exporter just work upon a backend relation or does it display metrics also when no connection possible (e.g. a `connection_down` metric)?

templates/prometheus-exporter.service.j2
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=prometheus exporter for pgbouncer
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/snap/bin/charmed-postgresql.prometheus-pgbouncer-exporter --web.listen-address=:{{ metrics_port }} --pgBouncer.connectionString="postgres://{{ stats_user }}:{{ stats_password }}@localhost:{{ listen_port }}/pgbouncer?sslmode=disable"
```
paulomach, line 7: the password will be written on the disk

for pg exporter, we used [snap configuration in the wrapper script](https://github.com/canonical/charmed-postgresql-snap/blob/d2260ebefb0af5eed273134158fc353099416b03/snap/local/start-exporter.sh#L18)

Does this approach make more sense?

paulomach (approved):
COS related is good, but I left some comments related to the pgb exporter itself

src/charm.py
dragomirp: > @dragomirp does the exporter just work upon a backend relation or does it display metrics also when no connection possible (e.g. a `connection_down` metric)?
It doesn't start without PGB running.

templates/prometheus-exporter.service.j2
taurus-forever: @dragomirp correct me, but different SNAPs will be installed ([parallel installs](https://snapcraft.io/docs/parallel-installs)) in case of several subordinate PGBs in the same principle charm. Isn't it?

dragomirp: > @dragomirp correct me, but different SNAPs will be installed ([parallel installs](https://snapcraft.io/docs/parallel-installs)) in case of several subordinate PGBs in the same principle charm. Isn't it?

In the future, yes, at the moment, no. Currently all the PGBs will share the same snap.

dragomirp: After a discussion, factored this out in a separate ticket [DPE-2042](https://warthogs.atlassian.net/browse/DPE-2042)

PR #78, opened 2023-06-05
[DPE-2049][DPE-2050] Interface limits and revision update

dragomirp:
## Issue
* The PGB charm will break if having multiple backend database relations
* The PGB charm should subordinate to only a single principle
* Snapped psql needs to make network connections

## Solution
* Limit interfaces where we want a one-to-one relationship
* Update to snap revision with fixed psql
* Remove a workaround for forcing the series of charms used for testing (fixed in libjuju)

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -53,6 +53,8 @@ requires:
   backend-database:
     interface: postgresql_client
     optional: false
+    limit: 1
   juju-info:
     interface: juju-info
     scope: container
+    limit: 1
--- a/src/constants.py
+++ b/src/constants.py
@@ -12,7 +12,7 @@ AUTH_FILE_NAME = "userlist.txt"
 # Snap constants.
 PGBOUNCER_EXECUTABLE = "charmed-postgresql.pgbouncer"
 POSTGRESQL_SNAP_NAME = "charmed-postgresql"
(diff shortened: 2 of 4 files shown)


metadata.yaml
dragomirp, line 1: I think if we remove the scope of `database`, `db` and `db-admin` we should be able to prevent PGB from subordinating to any charm, but then we'll need to relate to the principal twice (once for `juju-info` and once for the db relation).

taurus-forever: IMHO, the proper question: is `juju-info` still mandatory to be related for `subordinate` charms?
* yes => remove the scope of database, db and db-admin.
* no => why does it still exist?

taurus-forever (approved):

metadata.yaml
dragomirp: IIRC a container scoped requires interface is mandated on deploy.

dragomirp: The exact error:
```
$ juju deploy ./pgbouncer_ubuntu-22.04-amd64.charm
ERROR attempting to deploy "./pgbouncer_ubuntu-22.04-amd64.charm": subordinate charm "pgbouncer" lacks "requires" relation with container scope
```

marceloneppel (approved):

PR #88, opened 2023-06-19
[DPE-2047] Focal builds

dragomirp:

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -20,9 +20,16 @@ async def application_charm(ops_test: OpsTest):
 
 @pytest.mark.abort_on_fail
 @pytest.fixture(scope="module")
-async def pgb_charm(ops_test: OpsTest):
+async def pgb_charm_focal(ops_test: OpsTest):
     """Build the pgbouncer charm."""
     return await ops_test.build_charm(".", bases_index=0)
+
+
+@pytest.mark.abort_on_fail
+@pytest.fixture(scope="module")
+async def pgb_charm_jammy(ops_test: OpsTest):
+    """Build the pgbouncer charm."""
+    return await ops_test.build_charm(".", bases_index=1)
 
 
 @pytest.fixture(scope="module")
--- a/tests/integration/relations/test_db.py
+++ b/tests/integration/relations/test_db.py
@@ -25,11 +25,11 @@ RELATION_NAME = "db"
 
 
 @pytest.mark.abort_on_fail
-async def test_weebl_db(ops_test: OpsTest, pgb_charm) -> None:
+async def test_weebl_db(ops_test: OpsTest, pgb_charm_focal) -> None:
     """Deploy Weebl to test the 'db' relation."""
     backend_relation = await deploy_postgres_bundle(
         ops_test,
-        pgb_charm,
+        pgb_charm_focal,
         db_units=DATABASE_UNITS,
         pgb_series="jammy",
     )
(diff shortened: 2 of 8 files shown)


tests/integration/conftest.py
```diff
@@ -20,11 +20,18 @@ async def application_charm(ops_test: OpsTest):
...
     return await ops_test.build_charm(".", bases_index=0)
 
 
+@pytest.mark.abort_on_fail
+@pytest.fixture(scope="module")
+async def pgb_charm_jammy(ops_test: OpsTest):
+    """Build the pgbouncer charm."""
+    return await ops_test.build_charm(".", bases_index=1)
+
+
```
dragomirp, line 34: Depending on the test, we either need focal or jammy

tests/integration/relations/pgbouncer_provider/helpers.py
```diff
@@ -117,17 +118,20 @@ async def check_new_relation(ops_test: OpsTest, unit_name, relation_id, dbname):
...
+                f"SELECT data FROM {table_name} WHERE data = '{smoke_val}';"
+            )
+            run_update_query = await run_sql_on_application_charm(
+                ops_test,
+                unit_name=unit_name,
+                query=smoke_query,
+                dbname=dbname,
+                relation_id=relation_id,
+            )
+            assert smoke_val in json.loads(run_update_query["results"])[0]
```
dragomirp, line 137: This check frequently fails, both here and in the bundle tests.

tests/integration/relations/test_db.py
dragomirp, line 1: Mostly restoring the old mailman tests from before the switch to subordinate.

tests/integration/relations/test_db_admin.py
dragomirp, line 1: Retried relating with landscape, but it didn't work, so restored the test from before the switch to subordinate.

tests/integration/relations/test_db.py
```diff
@@ -25,32 +30,71 @@
 
 
 @pytest.mark.abort_on_fail
-async def test_weebl_db(ops_test: OpsTest, pgb_charm) -> None:
```
marceloneppel, line 28: Just out of curiosity: does weebl still work and relate correctly when deployed on focal?

tests/integration/relations/test_db_admin.py
marceloneppel: I forgot to take a look in the external services charm that was commented on https://chat.canonical.com/canonical/pl/68iepr5z7fdtpns5su34rseteo. I'm putting the link here for us to remember that this may or may not help on the issue.

marceloneppel (approved):

tests/integration/relations/test_db.py
dragomirp: > Just out of curiosity: does weebl still work and relate correctly when deployed on focal?

There was a release of it mid month that seems to break things for both Postgresql and Pgbouncer ([DPE-2159](https://warthogs.atlassian.net/browse/DPE-2159)). Old versions seemed to work with focal subordinate.

tests/integration/relations/pgbouncer_provider/helpers.py
taurus-forever: Is it because PB is not yet-started/relation user fully created OR the data in databas is not fully propagated (I saw it somewhere, user exist but password is not yet here, etc. I do not recall, was it mysql or postgresql)?

taurus-forever (approved):
Superb!

tests/integration/relations/pgbouncer_provider/helpers.py
dragomirp: Could be the databag, I've seen it fail with invalid password, could be we are setting/resetting the active status too soon. Created [DPE-2179](https://warthogs.atlassian.net/browse/DPE-2179) to investigate and fix.

PR #93, opened 2023-06-27
[DPE-2055] PGB extensions suport

dragomirp:
## Issue
Pgbouncer would block when a legacy relation requests  extensions even if they are enabled in the backend Postgresql DB

## Solution
* Check for extensions in the backend relation and block accordingly
* Updated snap charm lib
* Since the subordinate unit will be removed on broken relation, there's no need to recheck extensions blocking

### DIFF
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -85,7 +85,7 @@ Some example relation data is below. All values are examples, generated in a run
 """  # noqa: W505
 
 import logging
-from typing import Dict, Iterable
+from typing import Dict, Iterable, List
 
 from charms.pgbouncer_k8s.v0 import pgb
 from charms.postgresql_k8s.v0.postgresql import (
@@ -111,7 +111,7 @@ from ops.model import (
     WaitingStatus,
 )
 
-from constants import PG
+from constants import EXTENSIONS_BLOCKING_MESSAGE, PG
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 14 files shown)


taurus-forever (approved):
Nice!

tests/integration/relations/test_db.py
```diff
@@ -98,3 +101,45 @@ async def test_remove_relation(ops_test: OpsTest):
...
+        await ops_test.model.add_relation(f"{CLIENT_APP_NAME}:db", f"{pgb_jammy}:db")
+        # Pgbouncer enters a blocked status without a postgres backend database relation
+        await ops_test.model.wait_for_idle(apps=[PG], status="active", timeout=600)
+        await ops_test.model.wait_for_idle(apps=[pgb_jammy], status="blocked", timeout=600)
+        assert (
+            ops_test.model.units[f"{pgb_jammy}/0"].workload_status_message
+            == EXTENSIONS_BLOCKING_MESSAGE
+        )
+
+        logger.info("Rekatung with enabled extensions")
```
marceloneppel, line 131:
```suggestion
        logger.info("Relating with enabled extensions")
```

src/relations/db.py
```diff
@@ -165,6 +165,34 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
+
+    def _block_on_extensions(self, relation, remote_app_databag: Dict) -> bool:
+        """Verifies that extensions are enabled or blocks the charm."""
+        if "extensions" in remote_app_databag:
+            self.update_databags(
+                relation,
+                {
+                    "extensions": remote_app_databag["extensions"],
+                },
+            )
```
marceloneppel, line 184: Do you need this for this charm? On https://github.com/canonical/pgbouncer-k8s-operator/pull/100/files#r1240060895 you commented that it was needed to clean up when breaking the relation, but it seems that is not need as per the PR description.

marceloneppel (approved):

PR #106, opened 2023-08-16
[DPE-1765] Juju 3 peer secrets

dragomirp:
## Issue
Charm doesn't use secrets for internal sensitive values

## Solution
* Use Juju secrets instead of peer data if available
* Updated charm libs

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -286,7 +286,7 @@ class PgBouncerCharm(CharmBase):
         if juju_version.has_secrets:
             return self._juju_secret_get_key(scope, key)
 
-    def _juju_secret_set(self, scope: str, key: str, value: str) -> str:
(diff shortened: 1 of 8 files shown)


src/charm.py
```diff
@@ -197,6 +213,175 @@ def version(self) -> str:
...
+                logging.debug(f"Key {scope}:{key} has this value defined already")
+            else:
+                secret_cache[key] = value
+                try:
+                    secret.set_content(secret_cache)
+                except OSError as error:
+                    logging.error(
+                        f"Error in attempt to set {scope}:{key}. "
+                        f"Existing keys were: {list(secret_cache.keys())}. {error}"
+                    )
```
marceloneppel, line 316: Should we return here to avoid logging the debug message after this point?

marceloneppel (approved):

taurus-forever (approved):
LGTM after improving Marcelo finding.

PR #107, opened 2023-08-17
[MISC] Switch to Ruff

dragomirp:
## Issue
Charmcraft templates default to ruff as a linter and keeping flake8 depends on some packages that appear unmaintained

## Solution
Switch to ruff while maintaining compatibility with preexisting checks


src/relations/db.py
```diff
@@ -14,13 +14,13 @@
 the new postgres charm.
 
 Some example relation data is below. All values are examples, generated in a running test instance.
-┏━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
+------------------------------------------------------------------------------------------------------------
```
dragomirp, line 17: Ruff doesn't like the horizontal lines

PR #120, opened 2023-09-19
[MISC] Fix backend tests

dragomirp:
## Issue
Postgresql log location changed.

## Solution
Update the test to look at the right place for TLS connection.

### DIFF
--- a/tests/integration/relations/test_backend_database.py
+++ b/tests/integration/relations/test_backend_database.py
@@ -112,11 +112,10 @@ async def test_tls_encrypted_connection_to_postgres(ops_test: OpsTest, pgb_charm
 
         # Check the logs to ensure TLS is being used by PgBouncer.
         postgresql_primary_unit = await get_postgres_primary(ops_test)
-        logs = await run_command_on_unit(
+        mailman_ssl_log = f"connection authorized: user={pgb_user} database=mailman3 SSL enabled"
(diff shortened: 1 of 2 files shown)


tests/integration/relations/test_backend_database.py
```diff
@@ -112,11 +112,10 @@ async def test_tls_encrypted_connection_to_postgres(ops_test: OpsTest, pgb_charm
...
         postgresql_primary_unit = await get_postgres_primary(ops_test)
-        logs = await run_command_on_unit(
+        mailman_ssl_log = f"connection authorized: user={pgb_user} database=mailman3 SSL enabled"
+        postgresql_logs = "/var/snap/charmed-postgresql/common/var/log/postgresql/postgresql-*.log"
+        await run_command_on_unit(
             ops_test,
             postgresql_primary_unit,
-            "journalctl -u snap.charmed-postgresql.patroni.service",
+            f"grep '{mailman_ssl_log}' {postgresql_logs}",
         )
```
shayancanonical, line 121: why remove the assertion? should we no longer check that `grep` actually returns something?

dragomirp: Because `run_command_on_unit()` will raise on non zero return code and grep should exit with 1 if it doesn't find anything.

marceloneppel (approved):

taurus-forever (approved):

PR #129, opened 2023-10-17
[DPE-1770] Minor version upgrades

dragomirp:
Minor version upgrades.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -86,7 +86,7 @@ jobs:
           - false
         include:
           - tox-environments: client-relation-integration
-            agent-versions: "2.9.44"  # renovate: latest juju 2
+            agent-versions: "2.9.45"  # renovate: latest juju 2
             free-disk-space: true
           - tox-environments: client-relation-integration
             agent-versions: "3.1.6"  # renovate: latest juju 3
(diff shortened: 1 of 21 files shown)


.github/workflows/ci.yaml
```diff
@@ -135,9 +136,9 @@ jobs:
             echo "mark_expression=" >> $GITHUB_OUTPUT
           else
             echo Skipping unstable tests
-            echo "mark_expression=not unstable" >> $GITHUB_OUTPUT
+            echo "mark_expression=and not unstable" >> $GITHUB_OUTPUT
```
dragomirp, line 139: This should fix the scheduled tests.

src/charm.py
```diff
@@ -169,6 +148,39 @@ def _on_install(self, _) -> None:
...
+    def _on_install(self, _) -> None:
+        """On install hook.
+
+        This initialises local config files necessary for pgbouncer to run.
+        """
+        self.unit.status = MaintenanceStatus("Installing and configuring PgBouncer")
+
+        # Install the charmed PostgreSQL snap.
+        try:
+            self._install_snap_packages(packages=SNAP_PACKAGES)
```
taurus-forever, line 160: should we hold the snap?

taurus-forever (approved):
Nice job, just a one question otherwise LGTM!

src/charm.py
dragomirp: If it's installed by revision, it should be held in [`_install_snap_packages()`](https://github.com/canonical/pgbouncer-operator/blob/f331c218907afef40909df1837e5edd19df75654/src/charm.py#L665)

marceloneppel (approved):

PR #132, opened 2023-11-23
[DPE-3040] Update snap

dragomirp:
Update to pgbouncer 1.21

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -320,7 +320,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 22
+LIBPATCH = 24
 
 PYDEPS = ["ops>=2.0.0"]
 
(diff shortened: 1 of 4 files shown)


dragomirp:
The `--reboot`, the [docs](https://www.pgbouncer.org/usage.html#command-line-switches) recommend using rolling restart on the same socket. Should I try to get rid of it in this PR?

marceloneppel:
> The `--reboot`, the [docs](https://www.pgbouncer.org/usage.html#command-line-switches) recommend using rolling restart on the same socket. Should I try to get rid of it in this PR?

IMO, we can schedule it for the next pulse (with a new ticket) to avoid that it impacts the remaining tasks from this pulse (or just work on it after them if they are finished earlier).

marceloneppel (approved):

taurus-forever (approved):
Tnx!

taurus-forever:
> The `--reboot`, the [docs](https://www.pgbouncer.org/usage.html#command-line-switches) recommend using rolling restart on the same socket. Should I try to get rid of it in this PR?

IMHO, new ticket as it is a deprecation notification only. The `--reboot` should still do the job.
Sure we need to plane migration to the socket the next pulse.

dragomirp:
> > The `--reboot`, the [docs](https://www.pgbouncer.org/usage.html#command-line-switches) recommend using rolling restart on the same socket. Should I try to get rid of it in this PR?
> 
> IMHO, new ticket as it is a deprecation notification only. The `--reboot` should still do the job. Sure we need to plane migration to the socket the next pulse.

Created [DPE-3058](https://warthogs.atlassian.net/browse/DPE-3058) and [DPE-3059](https://warthogs.atlassian.net/browse/DPE-3059) to track this.

PR #136, opened 2024-01-17
[MISC] focal binary libs

dragomirp:
Remove binary python dependecies and build psycopg from source.


taurus-forever (approved):
Thank you!!!

marceloneppel (approved):

PR #150, opened 2024-02-09
[DPE-3535] Rerender config from secrets during upgrade

dragomirp:
Update rendered a blank config and relied on perpetual peer changed to refresh the configs.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -135,12 +135,12 @@ class PgBouncerCharm(CharmBase):
     #  Charm Lifecycle Hooks
     # =======================
 
-    def render_utility_files(self):
+    def render_utility_files(self, cfg_file=None):
         """Render charm utility services and configuration."""
         # Initialise pgbouncer.ini config files from defaults set in charm lib and current config.
         # We'll add basic configs for now even if this unit isn't a leader, so systemd doesn't
         # throw a fit.
-        if not (cfg_file := self.get_secret(APP_SCOPE, CFG_FILE_DATABAG_KEY)):
+        if not cfg_file:
             cfg_file = pgb.DEFAULT_CONFIG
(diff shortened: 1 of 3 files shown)


taurus-forever (approved):
black magic, but LGTM. :-D
Will retest candidate->edge upgrade after review/merging edge.

src/upgrade.py
```diff
@@ -83,7 +84,8 @@ def _on_upgrade_granted(self, event: UpgradeGrantedEvent) -> None:
         self.charm._install_snap_packages(packages=SNAP_PACKAGES, refresh=True)
 
         self.charm.unit.status = MaintenanceStatus("restarting services")
-        self.charm.render_utility_files()
+        cfg = self.charm.get_secret(APP_SCOPE, CFG_FILE_DATABAG_KEY)
+        self.charm.render_utility_files(cfg)
```
dragomirp, line 88: We need the cfg from the secret, since otherwise the follower units will get stuck during refresh.

src/charm.py
```diff
@@ -135,12 +135,14 @@ def __init__(self, *args):
...
 
-    def render_utility_files(self):
+    def render_utility_files(self, cfg_file=None):
         """Render charm utility services and configuration."""
         # Initialise pgbouncer.ini config files from defaults set in charm lib and current config.
         # We'll add basic configs for now even if this unit isn't a leader, so systemd doesn't
         # throw a fit.
-        cfg = pgb.PgbConfig(pgb.DEFAULT_CONFIG)
+        if not cfg_file:
+            cfg_file = pgb.DEFAULT_CONFIG
```
dragomirp, line 144: The peer rel is not yet around when the charm initialises, so we need to use the default

marceloneppel (approved):

PR #204, opened 2024-04-09
[MISC] Switch to Saas landscape

dragomirp:
## Issue
Legacy PG charm fails to deploy on juju 2.9.49 so we can't deploy the default self-hosted landscape bundle

## Solution
Switch to SaaS landscape


tests/integration/test_subordinates.py
```diff
@@ -30,64 +27,48 @@
...
             config={"token": github_secrets["UBUNTU_PRO_TOKEN"]},
+            channel="latest/edge",
+            num_units=0,
+        ),
+        ops_test.model.deploy(
+            LS_CLIENT,
+            config={
+                "account-name": github_secrets["LANDSCAPE_ACCOUNT_NAME"],
+                "registration-key": github_secrets["LANDSCAPE_REGISTRATION_KEY"],
+                "ppa": "ppa:landscape/self-hosted-beta",
```
marceloneppel, line 57: I remember I saw an older version of Landscae in this PPA, something like 23.03. Is it desired to have that?

marceloneppel (approved):

tests/integration/test_subordinates.py
dragomirp: IIRC it was recommended to @carlcsaposs-canonical from the Mysql charms. This is copied from there.

PR #210, opened 2024-04-13
[DPE-3927] Multiple databases

dragomirp:
* Adding wildcard database to be able to connect to multiple databases from PGB
* Adding limits to the number of PGB instances we spawn
* Adding Landscape test

There's still collection of read only DBs to do, in a follow up PR.


src/charm.py
```diff
@@ -598,6 +598,11 @@ def _get_relation_config(self) -> [Dict[str, Dict[str, Union[str, bool]]]]:
                     "port": r_port,
                     "auth_user": self.backend.auth_user,
                 }
+        pgb_dbs["*"] = {
+            "host": host,
+            "port": port,
+            "auth_user": self.backend.auth_user,
+        }
```
dragomirp, line 605: Wildcard for immediate RW access. DB collection TBD

templates/pgb_config.j2
```diff
@@ -1,9 +1,15 @@
 [databases]
-{% for name, database in databases.items() %}
-{{ name }} = host={{ database.host }} dbname={{ database.dbname }} port={{ database.port }} auth_user={{ database.auth_user }}
+{% for name, database in databases.items() -%}
+{{ name }} = host={{ database.host }} {% if database.dbname %}dbname={{ database.dbname }}{% else %}auth_dbname=postgres{% endif %} port={{ database.port }} auth_user={{ database.auth_user }}
+{% endfor %}
+
+[peers]
+{% for peer in peers -%}
+{{ peer + 1 }} = host={{ base_socket_dir }}{{ peer }} port={{ listen_port }}
```
dragomirp, line 8: Adding [peers](https://www.pgbouncer.org/config.html#section-peers). Not strictly necessary, but might as well.

src/relations/db.py
```diff
@@ -219,12 +219,10 @@ def _on_relation_joined(self, join_event: RelationJoinedEvent):
...
-            # If there's nothing in either databag, return early.
-            no_db = "No database name provided in app or unit databag"
-            logger.warning(no_db)
-            self.charm.unit.status = WaitingStatus(no_db)
-            join_event.defer()
-            return
+            # Sometimes a relation changed event is triggered, and it doesn't have
+            # a database name in it (like the relation with Landscape server charm),
+            # so create a database with the other application name.
+            database = join_event.relation.app.name
```
dragomirp, line 225: Landscape offers no db name.

src/relations/db.py
```diff
@@ -241,14 +239,13 @@ def _on_relation_joined(self, join_event: RelationJoinedEvent):
...
-                "database": database,
-            },
-        )
+        creds = {
+            "user": user,
+            "password": password,
+            "database": database,
+        }
+        self.charm.peers.app_databag[user] = json.dumps(creds)
+        self.update_databags(join_event.relation, creds)
```
dragomirp, line 248: Subordinate relations only show the principal unit and data. Landscape errors out if I try to use the app databag, the leader can't set this in the follower units (since it can't see them) and there's nowhere to read it from for the followers to set up themselves. Storing in the clear, since legacy relation credentials are also not using secrets.

templates/pgb_config.j2
```diff
@@ -1,9 +1,15 @@
 [databases]
-{% for name, database in databases.items() %}
-{{ name }} = host={{ database.host }} dbname={{ database.dbname }} port={{ database.port }} auth_user={{ database.auth_user }}
+{% for name, database in databases.items() -%}
+{{ name }} = host={{ database.host }} {% if database.dbname %}dbname={{ database.dbname }}{% else %}auth_dbname=postgres{% endif %} port={{ database.port }} auth_user={{ database.auth_user }}
```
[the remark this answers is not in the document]
dragomirp: I tried with `auth_dbname=pgbouncer` locally and it didn't work out, I think pgb might be trying to use its virtual table (for stats etc.) will investigate further. If it doesn't work with the pgbouncer db, I can try with the first db from the relation mapping, that should have already been setup.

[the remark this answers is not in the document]
marceloneppel: I think it's okay for the VM charm to use the first database from the relation mapping, as we have only one app related to PgBouncer. As for the K8S charm, we don't have such a limitation, allowing multiple applications. We should think better about which database to use in that case.

src/relations/pgbouncer_provider.py
```diff
@@ -152,6 +152,9 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
 
         dbs = self.charm.generate_relation_databases()
         dbs[str(event.relation.id)] = {"name": database, "legacy": False}
+        roles = extra_user_roles.lower().split(",")
+        if "admin" in roles or "superuser" in roles:
+            dbs["*"] = {"name": "*", "auth_dbname": database}
```
dragomirp, line 157: Adding the wildcard db only when we are admin. The presumption here is that a regular relation user is going to use only the provided db, so there's no sense in wasting PG connections on pools that wouldn't be used. Thoughts?

src/charm.py
```diff
@@ -535,6 +535,7 @@ def generate_relation_databases(self) -> Dict[str, Dict[str, Union[str, bool]]]:
             return {}
 
         databases = {}
+        add_wildcard = False
```
dragomirp, line 538: Adding admin wildcard logic here just in case we want to remove the interface limits in the future. The wildcard DB will be injected together with the actual client DB when joining the rel.

src/charm.py
```diff
@@ -550,16 +551,22 @@ def generate_relation_databases(self) -> Dict[str, Dict[str, Union[str, bool]]]:
...
+            roles = data.get("extra-user-roles", "").lower().split(",")
             if database:
                 databases[str(rel_id)] = {
                     "name": database,
                     "legacy": False,
                 }
+            if "admin" in roles or "superuser" in roles:
+                add_wildcard = True
+        if add_wildcard:
+            databases["*"] = {"name": "*", "auth_dbname": database}
```
dragomirp, line 569: Any db with the auth function should do, but we should have only one anyway.

templates/pgb_config.j2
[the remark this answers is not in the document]
dragomirp: Updated to use a relation database.

src/relations/pgbouncer_provider.py
marceloneppel: Maybe also add `createdb`?

src/relations/db.py
taurus-forever: Sad. Please fire a backlog Landscape ticket to migrate from legacy to modern postgresql_client interface.

taurus-forever (approved):
Nice work! :crossed_fingers:

marceloneppel (approved):

src/relations/pgbouncer_provider.py
dragomirp: Added `createdb`

marceloneppel (approved):

lucasgameiroborges (approved):

taurus-forever:
@rajannpatel we are merging this PR in edge to continue moving forward and improve edge based on your feedback.

PR #207, opened 2024-04-10
data_platform_libs update

juditnovak:
## Issue

Some tests were executed with incorrect mocks. Thus we had "fake" green results. The latest `data-interfaces` changes highlight on that.

Also some tests (suffering of the above) were irrelevant for PGbouncer and could be removed.

PR #220, opened 2024-04-23
 [MISC] Update data-platform-libs: data_interfaces to 34

lucasgameiroborges:


tests/unit/test_charm.py
```diff
@@ -636,80 +619,73 @@ def test_delete_existing_password_secrets(self, _):
...
-    @parameterized.expand([("app", True), ("unit", True), ("unit", False)])
+    @parameterized.expand([
+        ("app", True, "monitoring-password"),
+        ("unit", True, "key"),
+        ("unit", False, "key"),
+    ])
     @patch_network_get(private_address="1.1.1.1")
-    @patch("charm.JujuVersion.has_secrets", new_callable=PropertyMock, return_value=True)
-    def test_migration_from_single_secret(self, scope, is_leader, _):
+    @pytest.mark.usefixtures("with_juju_secrets")
```
lucasgameiroborges, line 664: Apparently, juju secrets are not enabled in this test with the old patch, which is why I had to use the new fixture here (and also on https://github.com/canonical/pgbouncer-k8s-operator/pull/277). Me and @shayancanonical investigated this issue, trying to figure out why the behavior is different, to no avail. If someone has an explanation to why that is, please let me know.

dragomirp:
Duplicate of https://github.com/canonical/pgbouncer-operator/pull/207

tests/unit/test_charm.py
juditnovak: It's been an issue I'm lobbying for from the start.

I stronlgy believe that ALL unittests must be run both with and without secrets.
I think on `mysql-router` we have a way to run them as such, in the same pipeline (allowing for coverage results collection -- which is normally the contra-argument why not to do it in the "easy way" in separate pieplines).

I can only ask for this once again... Guys _please_ run ALL your unittests **both** with and without secrets.
It's virtually impossible to follow, which of your worklfows may or may not use them. Since peer data is used pretty much everywhere.

The safe way (and reaslitic feedback for coverage) is to run all unittests under both "secret-ful" and "secret-less".

dragomirp:
Superseded by #207

PR #224, opened 2024-04-26
[DPE-4221] Recreate auth_query on backend rerelation

dragomirp:
## Issue
Auth query functions are not readded when backend rerelates

## Solution
Collect existing client rel dbs and add or remove auth_query function on backend rel events

### DIFF
--- a/src/relations/backend_database.py
+++ b/src/relations/backend_database.py
@@ -132,7 +132,9 @@ class BackendDatabaseRequires(Object):
                     logger.debug("database %s not yet created", database)
 
         for relation in self.charm.model.relations.get("db-admin", []):
-            database = self.legacy_db_admin_relation.get_databags(relation)[0].get("database")
+            database = self.charm.legacy_db_admin_relation.get_databags(relation)[0].get(
+                "database"
+            )
             if database and relation.units:
                 try:
                     con = self.postgres._connect_to_database(database)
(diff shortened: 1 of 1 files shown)


src/relations/backend_database.py
```diff
@@ -118,6 +118,43 @@ def __init__(self, charm: CharmBase):
             charm.on[BACKEND_RELATION_NAME].relation_broken, self._on_relation_broken
         )
 
+    def collect_databases(self) -> List[str]:
+        """Collects the names of all client dbs to inject or remove the auth_query."""
+        databases = [self.database.database, PG]
```
dragomirp, line 123: Kept, but I don't think we need to touch those dbs (`pgbouncer` and `postgres`)

src/relations/backend_database.py
```diff
@@ -118,6 +118,43 @@ def __init__(self, charm: CharmBase):
...
+        for relation in self.charm.model.relations.get("db", []):
+            database = self.charm.legacy_db_relation.get_databags(relation)[0].get("database")
+            if database and relation.units:
+                try:
+                    con = self.postgres._connect_to_database(database)
+                    con.close()
+                    databases.append(database)
+                except psycopg2.OperationalError:
+                    logger.debug("database %s not yet created", database)
```
dragomirp, line 132: Alternatively we can check for db existence when trying to add/remove the auth function

taurus-forever (approved):
Tested on Async-Cloud, works like a charm !

marceloneppel (approved):

delgod (approved):

PR #228, opened 2024-05-10
[DPE-3927] Collect readonly dbs

dragomirp:
Collect read only dbs from the backend postgresql when wildcard dbs are enabled.

Split off of https://github.com/canonical/pgbouncer-operator/pull/210 for easier review.

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -10,6 +10,3 @@ set-tls-private-key:
     private-key:
       type: string
       description: The content of private key for communications with clients. Content will be auto-generated if this option is not specified.
-
-update-readonly-dbs:
-  description: Update unrelated readonly databases in the backend Postgresql server.
--- a/actions.yaml
+++ b/actions.yaml
@@ -10,3 +10,6 @@ set-tls-private-key:
     private-key:
       type: string
       description: The content of private key for communications with clients. Content will be auto-generated if this option is not specified.
+
+update-readonly-dbs:
+  description: Update unrelated readonly databases in the backend Postgresql server.
--- a/src/charm.py
+++ b/src/charm.py
@@ -428,7 +428,7 @@ class PgBouncerCharm(CharmBase):
         return readonly_dbs
 
     def _collect_readonly_dbs(self) -> None:
-        if self.unit.is_leader() and self.backend.relation:
+        if self.unit.is_leader() and self.backend.postgres:
             existing_dbs = [db["name"] for db in self.get_relation_databases().values()]
             existing_dbs += ["postgres", "pgbouncer"]
             try:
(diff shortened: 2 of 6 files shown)


actions.yaml
```diff
@@ -10,3 +10,6 @@ set-tls-private-key:
     private-key:
       type: string
       description: The content of private key for communications with clients. Content will be auto-generated if this option is not specified.
+
+update-readonly-dbs:
+  description: Update unrelated readonly databases in the backend Postgresql server.
```
dragomirp, line 15: Action so the used doesn't need to wait for update status.

src/charm.py
```diff
@@ -401,6 +405,51 @@ def _on_start(self, _) -> None:
...
+                    r_port = r_host.split(":")[1]
+                    break
+
+                backend_databases = json.loads(self.peers.app_databag.get("readonly_dbs", "[]"))
+                for name in backend_databases:
+                    readonly_dbs[f"{name}_readonly"] = {
+                        "host": r_hosts,
+                        "dbname": name,
+                        "port": r_port,
+                        "auth_dbname": databases["*"]["auth_dbname"],
```
dragomirp, line 424: We need an auth_dbname, since the new readonly dbs don't have the query injected.

actions.yaml
taurus-forever: Actions must be approved by Mohamed and Mykola.
Maybe: https://github.com/PietroPasotti/resurrect ?

dragomirp: Got agreement in principle, but moving the action in a separate PR (https://github.com/canonical/pgbouncer-operator/pull/236) until everyone agrees on the specifics.

src/charm.py
taurus-forever: Is it a time to discuss the idea to put auth_query into template1 and make it always available?

templates/pgb_config.j2
```diff
@@ -2,6 +2,9 @@
 {% for name, database in databases.items() -%}
 {{ name }} = host={{ database.host }} {% if database.dbname %}dbname={{ database.dbname }}{% else %}auth_dbname={{ database.auth_dbname }}{% endif %} port={{ database.port }} auth_user={{ database.auth_user }}
 {% endfor %}
+{% for name, database in readonly_databases.items() -%}
+{{ name }} = host={{ database.host }} dbname={{ database.dbname }} auth_dbname={{ database.auth_dbname }} port={{ database.port }} auth_user={{ database.auth_user }}
+{% endfor %}
```
taurus-forever, line 7: I would be great to make sure PGB is OK having 100 lines in the config.

taurus-forever (approved):
LGTM.

marceloneppel (approved):

src/charm.py
marceloneppel: I suspect that even with that we may lack some permission to execute the query due to the following statement from the [PostgreSQL documentation](https://www.postgresql.org/docs/14/manage-ag-templatedbs.html):

> However, CREATE DATABASE does not copy database-level GRANT permissions attached to the source database. The new database has default database-level permissions.

PR #251, opened 2024-06-12
[DPE-4414][DPE-4450] Limit intances based on connection type

dragomirp:
* Only start multiple instances when exposing the charm
* Fix missing port with multiple units
* Bump charm libs
* Reloading the service


src/relations/pgbouncer_provider.py
```diff
@@ -111,14 +110,6 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
...
             return
 
-        # If exposed relation, open the listen port for all units
-        if event.external_node_connectivity:
-            # Open port
-            try:
-                self.charm.unit.open_port("tcp", self.charm.config["listen_port"])
-            except ModelError:
-                logger.exception("failed to open port")
-
```
dragomirp, line 121: The event is run only on the leader, so the other units didn't get the port opened.

src/charm.py
```diff
@@ -692,6 +703,17 @@ def render_pgb_config(self, reload_pgbouncer=False):
...
+        if self._is_exposed:
+            # Open port
+            try:
+                self.unit.open_port("tcp", self.config["listen_port"])
+            except ModelError:
+                logger.exception("failed to open port")
+
+            self.create_instance_directories()
+            self.render_utility_files()
+
```
dragomirp, line 716: When exposed, the charm should open the port and create config dirs for more instances. This will run more frequently than necessary, but I'm not sure there's a better place.

src/charm.py
```diff
@@ -159,11 +177,6 @@ def tracing_endpoint(self) -> Optional[str]:
 
     def render_utility_files(self):
         """Render charm utility services and configuration."""
-        # Initialise pgbouncer.ini config files from defaults set in charm lib and current config.
-        # We'll add basic configs for now even if this unit isn't a leader, so systemd doesn't
-        # throw a fit.
-        self.render_pgb_config()
-
```
dragomirp, line 166: Moved out so that `render_utility_files` can be called from `render_pgb_config`

src/relations/pgbouncer_provider.py
```diff
@@ -225,7 +213,7 @@ def update_connection_info(self, relation):
         if exposed:
             rw_endpoint = f"{self.charm.leader_ip}:{self.charm.config['listen_port']}"
         else:
-            rw_endpoint = f"localhost:{self.charm.config['listen_port']}"
+            rw_endpoint = f"{SOCKET_LOCATION}:{self.charm.config['listen_port']}"
```
dragomirp, line 216: This works for the test app, but may cause issues elsewhere? Should I keep this or switch to socket later on?

dragomirp: Both legacy test failed if I set the `host` to the socket location, so leaving the legacy interface be for the time being.

taurus-forever: I am lost, why `:{self.charm.config['listen_port']}` stays here "/path/socket:port" ? Should be socket only.
 
IMHO:
 * legacy interfaces: we must follow legacy logic (principal pgbouncer => ip+port).
 * modern interface: 
   - vm/subordinate/non-exposed = socket only (one PGB process).
   - k8s/principal/exposed = ip:port
Nope?

tests/unit/relations/test_pgbouncer_provider.py
```diff
@@ -108,7 +114,7 @@ def test_on_database_requested(
         _dbp_set_credentials.assert_called_with(rel_id, user, _password())
         _dbp_set_version.assert_called_with(rel_id, _pg().get_postgresql_version())
         _dbp_set_endpoints.assert_called_with(
-            rel_id, f"localhost:{self.charm.config['listen_port']}"
+            rel_id, f"{SOCKET_LOCATION}:{self.charm.config['listen_port']}"
```
taurus-forever, line 117: The same question as above.

taurus-forever (commented):
LGTM in general, all as we discussed.
One implementation question insidel.

src/relations/pgbouncer_provider.py
dragomirp: > I am lost, why `:{self.charm.config['listen_port']}` stays here "/path/socket:port" ? Should be socket only.

The test app splits the host and port [here](https://github.com/canonical/postgresql-test-app/blob/main/src/charm.py#L247). That's controlled by us, so easy to change, but it raises the question what other clients expect. I checked the maas-region charm and it's [passing the endpoint directly](https://github.com/canonical/maas-charms/blob/main/maas-region/src/charm.py#L131). I think some clients expect the socket directory and the port to connect (psql? I'll check locally).

dragomirp: Maybe add the socket endpoint once we have the URI in the interface? That might make things more straightforward to use.

marceloneppel (approved):

src/relations/pgbouncer_provider.py
dragomirp: Looks like maas-region is also a snapped application so we will likely need to offer a socket slot for everything to work. Reverted to localhost for the time being.

taurus-forever (approved):

src/charm.py
```diff
@@ -147,10 +146,29 @@ def __init__(self, *args):
...
     # =======================
 
+    def create_instance_directories(self):
+        """Create configuration directories for pgbouncer instances."""
+        pg_user = pwd.getpwnam(PG_USER)
+        app_conf_dir = f"{PGB_CONF_DIR}/{self.app.name}"
+
+        # Make a directory for each service to store configs.
+        for service_id in self.service_ids:
+            logger.error(f"{app_conf_dir}/{INSTANCE_DIR}{service_id}")
```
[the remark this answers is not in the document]
dragomirp: Yes.

marceloneppel (approved):

PR #254, opened 2024-06-17
[MISC] Don't fail when collecting readonly dbs

dragomirp:
Exit read only db collection if unable to connect to the backend database.

Example failure: https://github.com/canonical/postgresql-bundle/actions/runs/9537877087/job/26286454688


taurus-forever (approved):
Nice one-liner catch!

marceloneppel (approved):

Issue #245, opened 2024-06-10
pgbouncer refuses to authenticate the application under which it is subordinate


skatsaounis:
## Steps to reproduce

```bash
1. Install postgresql `14/stable`, pgbouncer `1/edge` or `1/stable` with the default settings and relate them with a single Juju Terraform run.
2. Install maas-region latest/edge charm and relate it to pgbouncer.
```

## Expected behavior

The maas-region charm receives the connection details through the `dabatase` relation and tries to run `maas init region+rack --database-uri postgres://relation_id_9:HZjrAAmo5U0i37DVwUEba7FB@localhost:6432/maas_region_db`. Details are coming through the relation. MAAS is initialized successfully.

## Actual behavior

MAAS fails to initialize since when it receives the connection details and tries to authenticate a refusal error occurs (see logs).

## Versions

Operating system: Ubuntu 22.04.4 LTS

Juju CLI: 3.4.3-genericlinux-amd64

Juju agent: 3.4.3

postgresql charm revision: 363
pgbouncer charm revision: 203

LXD: 5.0.3

## Log output

```bash
2024-06-10 12:40:01 INFO juju.worker.uniter.operation runhook.go:186 ran "api-relation-changed" hook (via hook dispatching script: dispatch)
2024-06-10 12:40:06 INFO juju.worker.uniter.operation runhook.go:186 ran "maas-db-relation-joined" hook (via hook dispatching script: dispatch)
2024-06-10 12:40:09 INFO juju.worker.uniter.operation runhook.go:186 ran "maas-db-relation-changed" hook (via hook dispatching script: dispatch)
2024-06-10 12:40:17 INFO juju.worker.uniter.operation runhook.go:186 ran "maas-db-relation-changed" hook (via hook dispatching script: dispatch)
2024-06-10 12:40:22 INFO unit.maas-region/0.juju-log server.go:325 maas-db:9: endpoints changed on 2024-06-10 12:40:22.992699
2024-06-10 12:40:23 INFO unit.maas-region/0.juju-log server.go:325 maas-db:9: MAAS database endpoints have been changed to: localhost:6432
2024-06-10 12:40:23 INFO juju.worker.uniter.operation runhook.go:186 ran "maas-db-relation-changed" hook (via hook dispatching script: dispatch)
2024-06-10 12:40:25 INFO unit.maas-region/0.juju-log server.go:325 maas-db:9: database created at 2024-06-10 12:40:25.993849
[...]

PR #257, opened 2024-06-19
[DPE-4595] Immediate connection on database created

dragomirp:
## Issue
Pgbouncer is not set up to receive connections immediately when publishing credentials

## Solution
Sync up configuration and reload PGB when establishing the interface

Related to https://github.com/canonical/pgbouncer-operator/issues/245


src/relations/pgbouncer_provider.py
```diff
@@ -110,6 +122,13 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
             event.defer()
             return
 
+        for key in self.charm.peers.relation.data.keys():
+            if key != self.charm.app:
+                if self.charm.peers.relation.data[key].get("auth_file_set", "false") != "true":
+                    logger.info("Backend credentials not yet set on all units")
+                    event.defer()
+                    return
+
```
dragomirp, line 131: Make sure all the units have the userlist rendered.

src/relations/pgbouncer_provider.py
```diff
@@ -91,13 +91,25 @@ def __init__(self, charm: CharmBase, relation_name: str = CLIENT_RELATION_NAME)
...
 
+    def _on_relation_changed(self, event: RelationChangedEvent) -> None:
+        """Injects the database name for follower units."""
+        if not self.charm.unit.is_leader():
+            if "database" in event.relation.data[event.app]:
+                database = event.relation.data[event.app]["database"]
+                dbs = self.charm.generate_relation_databases()
+                if database not in [db["name"] for db in dbs]:
+                    self.charm.render_pgb_config(reload_pgbouncer=True, inject_db=database)
+
```
dragomirp, line 112: Bypassing the lib to inject the requested database in follower pgbouncers. This us needed because `_on_database_requested` will only fire on the leader and followers will need to wait for peer data hooks to fire to catch up. Alternatively, I can try to checksum the dblists in peer data and defer `_on_database_requested` until all units are in sync. Thoughts?

dragomirp:
This seems to cause instability on juju 2 scale up, but I want early feedback since the changes are aggressive.

src/relations/pgbouncer_provider.py
```diff
@@ -148,6 +167,8 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
             dbs["*"] = {"name": "*", "auth_dbname": database}
         self.charm.set_relation_databases(dbs)
 
+        self.charm.render_pgb_config(reload_pgbouncer=True)
+
```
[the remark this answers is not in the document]
taurus-forever: > ... but it won't work if we have multiple client units.

It is subordinate... 1:1 relation here. Nope?

taurus-forever (approved):
LGTM, one point.

src/relations/pgbouncer_provider.py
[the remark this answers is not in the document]
dragomirp: PGB will be 1:1 for the client units, but it's not a given that we'll have a single client unit. Say we have a client with two units set up like so: c0*→pgb0 and c1→pgb1*, when establishing the relation pgb1 will create the user and db, update the db mapping, reload pgb and trigger a `database_created` event for the client. While pgb1 may be good to go, at the time `database_created` triggers, pgb0 may not yet be fully set up, it may not have synced up the dbs from peer data, it may still be setuping the backend relation. If c0 tries to use pgb0 while it's still setting up, the connection will error out.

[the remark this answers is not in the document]
marceloneppel: Is this fix related to https://github.com/canonical/pgbouncer-operator/issues/245?

src/relations/pgbouncer_provider.py
marceloneppel: IMO, it's better to wait for all units to be in sync because this event can still be in the follower unit queue (not executed yet) while the leader has already set the connection info in the client relation data.

marceloneppel: Another important thing for the current implementation of this PR, in my opinion, is to ensure that other calls to `render_pgb_config` in the follower units are safe and don't remove the injected database after it was added at this point. I mean the situation where this call injects the database, another queued event calls `render_pgb_config`, which removes the injected database, and later on, the follower unit receives the `peer-relation-changed` event (because the leader took some time to set the peer relation data).

src/relations/pgbouncer_provider.py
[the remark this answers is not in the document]
dragomirp: > Is this fix related to #245?

Yes

src/relations/pgbouncer_provider.py
dragomirp: I'll give it a go to sync configs.

dragomirp: Reworked to hash db list and defer until all units are in sync.

taurus-forever (approved):
640KB^w 16 digits will be enough for everyone!

LGTM.

marceloneppel (approved):
LGTM! Thanks, Dragomir!

PR #274, opened 2024-07-03
[DPE-4772] Subordinate scale up

dragomirp:
Implements a synchronisation mechanism so that scaled up units will only trigger `database_created` when pgbouncer is ready to accept connections.


src/relations/peers.py
```diff
@@ -143,6 +143,10 @@ def _on_changed(self, event: HookEvent):
         )
         self.charm.render_pgb_config(reload_pgbouncer=True)
         self.unit_databag["pgb_dbs"] = pgb_dbs_hash
+        if self.unit_databag.get("auth_file_set", "false") == "true" and self.unit_databag.get(
+            "backend_relation"
+        ):
+            self.charm.client_relation.set_ready()
```
dragomirp, line 149: At this point we should have a backend relation, rendered user list and config, so unit should be ready

src/relations/pgbouncer_provider.py
```diff
@@ -99,6 +107,9 @@ def _depart_flag(self, relation):
     def _unit_departing(self, relation):
         return self.charm.peers.unit_databag.get(self._depart_flag(relation), None) == "true"
 
+    def _on_relation_created(self, event: RelationCreatedEvent) -> None:
+        event.relation.data[self.charm.unit].update({"state": "setup"})
```
dragomirp, line 111: Exact value doesn't really matter as long as it's not `ready`

src/relations/pgbouncer_provider.py
```diff
@@ -168,6 +179,7 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
             return
 
         self.charm.render_pgb_config(reload_pgbouncer=True)
+        self.set_ready()
```
dragomirp, line 182: Setting the leader ready immediately, so we don't have to wait in a single unit cluster.

src/relations/backend_database.py
```diff
@@ -181,6 +181,7 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
             self.charm.render_pgb_config(reload_pgbouncer=True)
             self.charm.render_prometheus_service()
             self.charm.update_status()
+            self.charm.peers.unit_databag.update({"backend_relation": str(self.relation.id)})
```
dragomirp, line 184: We can most probably remove some of the logic here to peer changed, but let's cleanup after we have the general behaviour right.

src/relations/pgbouncer_provider.py
marceloneppel: Testing with a patched `postgresql-test-app` charm (with the same changes in the `data_interfaces` library we have here), I found that this hook may fire after the client application receives the relation changed event.

Steps to reproduce:
```sh
juju model-config logging-config="<root>=INFO;unit=DEBUG"
juju deploy postgresql --channel 14/stable
cd ~/pgbouncer-operator
tox -e build-production
juju deploy ./pgbouncer_ubuntu-22.04-amd64.charm
cd ~/postgresql-test-app # It must contain the same changes in the `data_interfaces` library.
tox -e build-production
juju deploy ./postgresql-test-app_ubuntu-22.04-amd64.charm
juju relate pgbouncer postgresql-test-app:first-database
juju relate pgbouncer postgresql
```

Wait for the deployment to settle, then run the following commands:
```sh
juju add-unit postgresql-test-app
juju debug-log -m lxd:dev --include unit-postgresql-test-app-1 --include unit-pgbouncer-1 --replay
```

In the debug logs, you can see that the relation-changed event fires in the client application earlier than the relation-created event in the PgBouncer unit.
```sh
…
unit-postgresql-test-app-1: 20:24:39 INFO juju.worker.uniter.operation ran "first-database-relation-created" hook (via hook dispatching script: dispatch)
…
unit-postgresql-test-app-1: 20:24:42 DEBUG unit.postgresql-test-app/1.juju-log first-database:6: No provider units are available.
unit-postgresql-test-app-1: 20:24:42 INFO juju.worker.uniter.operation ran "first-database-relation-changed" hook (via hook dispatching script: dispatch)
…
unit-postgresql-test-app-1: 20:25:07 INFO unit.postgresql-test-app/1.juju-log first-database:6: database created at 2024-07-04 20:25:07.984376
…
unit-postgresql-test-app-1: 20:25:08 INFO juju.worker.uniter.operation ran "first-database-relation-changed" hook (via hook dispatching script: dispatch)
…
unit-pgbouncer-1: 20:25:10 INFO juju.worker.uniter.operation ran "database-relation-created" hook (via hook dispatching script: dispatch)
…
unit-postgresql-test-app-1: 20:25:10 DEBUG unit.postgresql-test-app/1.juju-log first-database:6: Subordinate provider unit not ready.
…
unit-pgbouncer-1: 20:25:20 INFO unit.pgbouncer/1.juju-log backend-database:7: database created at 2024-07-04 20:25:20.936335
…
```

The status logs also show something that confirms that (`juju show-status-log postgresql-test-app/1 -n 1000`):
```sh
…
04 Jul 2024 20:25:07Z  juju-unit  executing    running first-database-relation-changed hook for pgbouncer/1
04 Jul 2024 20:25:08Z  workload   active       received database credentials of the first database
…
```
And for PgBouncer (`juju show-status-log pgbouncer/1 -n 1000`):
```sh
…
04 Jul 2024 20:25:09Z  juju-unit  executing    running database-relation-created hook
…
```

src/relations/pgbouncer_provider.py
```diff
@@ -107,6 +111,9 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
         Deferrals:
             - If backend relation is not fully initialised
         """
+        rel_id = event.relation.id
+        self.database_provides.set_subordinated(rel_id)
```
dragomirp, line 115: Set the app level flag. There's other synchronisation for existing units, so it should be safe even if not immediately set.

src/relations/backend_database.py
```diff
@@ -181,6 +181,7 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
             self.charm.render_pgb_config(reload_pgbouncer=True)
             self.charm.render_prometheus_service()
             self.charm.update_status()
+            self.charm.client_relation.set_ready()
```
dragomirp, line 184: We should have the peer data and secret here already, so no need to retrigger the peer changed hook.

marceloneppel (approved):
LGTM!

lucasgameiroborges (approved):

delgod (changes requested):

src/relations/pgbouncer_provider.py
```diff
@@ -251,6 +258,11 @@ def update_connection_info(self, relation):
         self.charm.unit.status = initial_status
         self.charm.update_status()
 
+    def set_ready(self) -> None:
+        """Marks the unit as ready for all database relations."""
+        for relation in self.model.relations[self.relation_name]:
+            relation.data[self.charm.unit].update({"state": "ready"})
```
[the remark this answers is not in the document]
dragomirp: `update_relation_data()` should be setting app level data, we need this in the unit databag. We'll need another helper for that.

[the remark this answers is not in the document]
dragomirp: Possibly I can use [_update_relation_data_without_secrets()](https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/data_interfaces.py#L1135-L1143) but that's private.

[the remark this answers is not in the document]
juditnovak: No, we can't use any of those. Cross-charm relations (incl. all related internal functions) are restricted to `app` data.

Only peer relations support interface functions to access unit data at the moment.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2842,6 +2852,24 @@ def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
 
     def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
         """Event emitted when the database relation has changed."""
+        is_subordinate = False
+        remote_unit_data = None
+        for key in event.relation.data.keys():
+            if isinstance(key, Unit) and not key.name.startswith(self.charm.app.name):
+                remote_unit_data = event.relation.data[key]
+            elif isinstance(key, Application) and key.name != self.charm.app.name:
+                is_subordinate = event.relation.data[key].get("subordinated") == "true"
```
juditnovak, line 2861: Huh, this feels kinda hacky... I get the rationale, but I think we should have a proper "frame" around this.

The `DatabaseProviderEventHander` -- up to this point was not supposed to directly dig in databags.

We are reading an manipulating relation data via the `DatabaseProviderEventHander.relation_data`... Which --as for now-- strictly mean `app` level data.

We can alter the existing workflows to be ready to handle unit level data as well.
However that should go alongside considerations similar to the existing ones across the module (not just direct access in various parts of the code).

I could go with this solution as a temporary one, and I'm glad to volunteer to introduce the missing concept of handling unit level data in the libs. At that point this code will be cleaned up.

dragomirp: Let's keep it as is for the time being.

delgod (approved):

PR #340, opened 2024-08-22
fix: ignore search_path startup parameter

skatsaounis:
## Issue

#341

## Solution

Ignore `search_path` startup parameter to allow connections that are using it to pass


dragomirp:
Thanks for the contribution. @taurus-forever we should most probably expose `ignore_startup_parameters` in the configuration, like the old charm.

dragomirp (approved):

taurus-forever (approved):

taurus-forever:
re: most probably expose ignore_startup_parameters in the configuration, like the old charm. LGTM.
@dragomirp prepare PR and include @delgod please. Tnx!

delgod (changes requested):
I don't think we should pass connections which don't work correctly.
search_path isn't supported by PgBouncer, and clients who use it may work incorrectly.
I think the right way is to change the client side to make it work without search_path.
In any case, the decision is up to the Product Manager.

taurus-forever:
Dear @skatsaounis , thank you for PR!  The review is still in draft, you need to press "Ready for the review".

In the same time. Can you please explain your usecase and need for `search_path`? Thank you!
BTW, are you using the connection uri provided by data-platform-libs or constructed the one yourself (can you please share it)?

Thank you!

skatsaounis:
Hi all and thank you for the review. This PR is wrong and as such I am planning to close it. My initial understanding of the `ignore_startup_parameters` was that pgbouncer ignores them but forwards them to postgres to handle them. As it seems, it is removing them from the connections entirely. It is true that `search_path` is not supported by pgbouncer, but if an application is relying on it, then the application should not use pgbouncer. Or it should be refactored to remove the need of `search_path` in order to use pgbouncer. Merging this PR would lead to erroneous situations where e.g. tables are wrongly created to public schema or similar.

PR #343, opened 2024-08-23
[DPE-4683] socket config flag

dragomirp:
* Add configuration flag to use socket
* Pass socket endpoint and URI based on the config flag


config.yaml
```diff
@@ -21,6 +21,12 @@ options:
       Virtual IP to use to front pgbouncer units. Used only in case of external node connection.
     type: string
 
+  local_connection_type:
+    default: localhost
+    description: |
+      Type of local connectivity to provide. Must be either 'localhost' or 'socket'. Only applies to the database interface.
+    type: string
```
dragomirp, line 28: vIP value will already control how exposed connections will behave, so adding config just for local connections.

marceloneppel (approved):

7annaba3l:
Let's rather have as possible values:
"tcp" instead of "socket"
"uds" instead of "local"

config.yaml
```diff
@@ -21,6 +21,12 @@ options:
       Virtual IP to use to front pgbouncer units. Used only in case of external node connection.
     type: string
 
+  local_connection_type:
+    default: tcp
+    description: |
+      Type of local connectivity to provide. Must be either 'tcp' or 'uds'. Only applies to the database interface.
+    type: string
+
```
taurus-forever, line 29: Did you sign this with Mohamed/Mykola/Jon (spec...). I am afraid it will be rejected.

P.S. I would use `tcp` or `socket` (instead of `uds`). Sure, 'the color of the button' story.

taurus-forever (commented):
Please sign naming with Mohamed. LGTM right after.

config.yaml
taurus-forever: Oh, just noticed https://github.com/canonical/pgbouncer-operator/pull/343#issuecomment-2334575517
@7annaba3l please sign the option name and values.

dragomirp: This should be the last piece missing for [DA117](https://docs.google.com/document/d/1cemtUHv9sMFf9S8x9AWgLJmYDToNFYCfd4QO1HdHGtU/edit?tab=t.0).

taurus-forever (approved):

marceloneppel (approved):

PR #369, opened 2024-09-23
[MISC] Disable cached builds

dragomirp:
* Temporarily disable cached builds until https://github.com/canonical/charmcraftcache-hub/issues/111 can be resolved
* Bump libs


taurus-forever (approved):
Sad, but we need to merge tested PRs (even slower).

marceloneppel (approved):

PR #373, opened 2024-09-27
[DPE-5564] Don't set secrets until db is set

dragomirp:
* Update libs
* Removes an OPS import warning
* Fixes wrong emission of `database_created` due to creating the secret before the requirer sets the requested database (data-integrator)

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -34,8 +34,8 @@ from ops import (
     Relation,
     StartEvent,
     WaitingStatus,
+    main,
 )
-from ops.main import main
 
 from config import CharmConfig
 from constants import (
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -34,8 +34,8 @@
     Relation,
     StartEvent,
     WaitingStatus,
+    main,
 )
-from ops.main import main
```
dragomirp, line 38: OPS complains for the import

tests/unit/relations/test_pgbouncer_provider.py
dragomirp, line 1: To circumvent early setting of data protection

lucasgameiroborges (approved):

taurus-forever (approved):

PR #384, opened 2024-10-08
[MISC] Stop tracking channel for held snaps

dragomirp:
Remove tracking of snap channel for the charmed-postgresql snap

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -12,12 +12,7 @@ AUTH_FILE_NAME = "userlist.txt"
 # Snap constants.
 PGBOUNCER_SNAP_NAME = "charmed-pgbouncer"
 PGBOUNCER_EXECUTABLE = f"{PGBOUNCER_SNAP_NAME}.pgbouncer"
-SNAP_PACKAGES = [
-    (
-        PGBOUNCER_SNAP_NAME,
-        {"revision": {"aarch64": "15", "x86_64": "16"}, "channel": "1/stable"},
-    )
-]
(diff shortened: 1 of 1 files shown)


marceloneppel (approved):

lucasgameiroborges (approved):

lucasgameiroborges:
Huh, just realized this is going to break the automation for [pinning snap revs in bundles](https://github.com/canonical/data-platform-workflows/blob/main/python/cli/data_platform_workflows_cli/update_bundle.py#L173-L183) :(

Question: Will the channel be specified anywhere else in the charm code? Will there be any other "source of truth" for the channel of the snap? Wondering if I can avoid hard-coding it in the workflow.

dragomirp:
> Question: Will the channel be specified anywhere else in the charm code? Will there be any other "source of truth" for the channel of the snap? Wondering if I can avoid hard-coding it in the workflow.

Technically the channel stated is wrong, there's no revisions in it ATM. We can set the correct risk (14/edge and 1/edge), but then we can't change it when releasing a charm revision to stable. Is the revision itself not enough for store pinning to work?

taurus-forever (approved):

taurus-forever:
> Huh, just realized this is going to break the automation for [pinning snap revs in bundles](https://github.com/canonical/data-platform-workflows/blob/main/python/cli/data_platform_workflows_cli/update_bundle.py#L173-L183) :(

Follow MySQL approach for now: https://github.com/canonical/data-platform-workflows/blob/main/python/cli/data_platform_workflows_cli/update_bundle.py#L121

PR #393, opened 2024-10-17
[DPE-5613] Enable round robin connections to read only backend nodes

dragomirp:
## Issue
PGB doesn't load balance the read only databases when multiple read only backend units are available

## Solution
Add `server_round_robin` to enable round robin connections when multiple hosts are available.


marceloneppel (approved):
LGTM!

taurus-forever (approved):

PR #421, opened 2024-11-28
[MISC] Python version

dragomirp:
## Issue
The cryptography package dropped support for python 3.7, but the new version specification fails poetry's dependency resolution: https://github.com/canonical/pgbouncer-operator/pull/419#issuecomment-2505092783

## Solution
Since cryptography 44 should still work on python 3.8 (focal), split the python dependency into 3.8 and greater that 3.9


pyproject.toml
```diff
@@ -5,14 +5,14 @@
 package-mode = false
 
 [tool.poetry.dependencies]
-python = "^3.8.10"
+python = ">=3.8,<3.9 || >=3.10,<4"
```
dragomirp, line 8: Cryptography doesn't resolve for 3.9. While jammy uses 3.10, I left the second directive loose so that we can still resolve dependencies on 3.12 (noble, rolling distros, etc.)

lucasgameiroborges (approved):

marceloneppel (approved):

PR #518, opened 2025-03-24
[DPE-6555] Fix unexposed and vIP read only URIs

dragomirp:
Readonly URIs are failing in some cases. Fixing known cases and adding unit test.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -29,7 +29,7 @@ parts:
       PIP_BREAK_SYSTEM_PACKAGES=true python3 -m pip install --user --upgrade pip==25.0.1  # renovate: charmcraft-pip-latest
 
       # Use uv to install poetry so that a newer version of Python can be installed if needed by poetry
-      curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.6.7/uv-installer.sh | sh  # renovate: charmcraft-uv-latest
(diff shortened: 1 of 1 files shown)


src/relations/pgbouncer_provider.py
```diff
@@ -354,21 +354,20 @@ def update_read_only_endpoints(
...
-                    relation.id, f"{self.charm.config.vip}:{port}"
-                )
-                read_only_uri = (
-                    f"postgresql://{user}:{password}@{self.charm.config.vip}:{port}/{database}_readonly",
-                )
+                if self.charm.config.vip:
+                    self.database_provides.set_read_only_endpoints(
+                        relation.id, f"{self.charm.config.vip}:{port}"
+                    )
+                    read_only_uri = f"postgresql://{user}:{password}@{self.charm.config.vip}:{port}/{database}_readonly"
```
dragomirp, line 361: vIP should be used when the relation is exposed.

src/relations/pgbouncer_provider.py
```diff
@@ -382,10 +381,6 @@ def update_read_only_endpoints(
             password = None
             database = None
 
-    def get_database(self, relation):
-        """Gets database name from relation."""
-        return relation.data.get(self.get_external_app(relation)).get("database", None)
-
```
dragomirp, line 388: Doesn't seem to be used anywhere.

marceloneppel (approved):

PR #540, opened 2025-04-29
[MISC] Switch to uv plugin

dragomirp:
Experimental switch to the uv parts plugin.


tox_uv.toml
dragomirp, line 1: Workaround file to disable build packages for tox envs. Once charmcraft starts building by default, we should be able to remove.

.github/workflows/ci.yaml
```diff
@@ -25,9 +25,23 @@ on:
         value: ${{ jobs.build.outputs.artifact-prefix }}
 
 jobs:
+  # TODO install uv in reusable and re-enable
+  # lint:
```
dragomirp, line 29: Reusable lint job doesn't have uv

.github/workflows/ci.yaml
```diff
@@ -50,6 +64,8 @@ jobs:
   build:
     name: Build charm
     uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v31.0.1
+    with:
+      cache: false
```
dragomirp, line 68: I'm not sure if cache will be picked up by uv, when re-enabled.

pyproject.toml
```diff
@@ -1,74 +1,70 @@
...
-[build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+[project]
+name = "pgbouncer-operator"
+version = "0.1.0"
+description = ""
+readme = "README.md"
+requires-python = ">=3.8"
+[dependency-groups]
```
dragomirp, line 10: Optional groups for everything, so we only install what we need.

pyproject.toml
```diff
@@ -1,74 +1,70 @@
...
+charm = [
+    "jinja2==3.1.6",
+    "ops==2.20.0",
+    "tenacity==9.0.0",
+    "pydantic==1.10.22",
+    "psycopg2==2.9.10",
+    "psycopg[c]==3.2.6"
```
dragomirp, line 17: Those will be rebuild in the unit tests as before. To add them as binaries for the tests without magic, we should have a solution for the no-binaries option.

pyproject.toml
```diff
@@ -1,74 +1,70 @@
...
+    "landscape-api-py3==0.9.0",
+    "allure-pytest==2.14.0",
+    "allure-pytest-default-results==0.1.2"
+]
+
+[tool.uv]
+no-binary = true
+build-constraint-dependencies = [
+    "setuptools_scm < 8.3.0; python_version < '3.10'"
+]
```
dragomirp, line 67: Force source builds and install the latest working version of setuptools_scm for python 3.8

spread.yaml
```diff
@@ -115,14 +115,15 @@ prepare: |
...
+  pipx install tox
 prepare-each: |
   cd "$SPREAD_PATH"
   if [[ $SPREAD_VARIANT == *"juju29"* ]]
   then
     # Each version of python-libjuju is only compatible with one major Juju version
     # Override python-libjuju version pinned in poetry.lock
-    poetry add --lock --group integration juju@^2
+    uv remove --group integration juju
+    uv --config-file=tox_uv.toml add --group integration 'juju<3'
```
dragomirp, line 126: Passing config file to have binary deps in tests, but source builds in the charm itself.

tox_uv.toml
dragomirp, line 1: Workaround for things where we don't want to build (linting and test dependencies). When the uv plugin starts rebuilding everything by default, we should be able to remove this.

tox.ini
dragomirp, line 1: There's a tox-uv plugin. If we decide to switch permanently, it might be worth it to try it.

.github/workflows/ci.yaml
```diff
@@ -25,9 +25,23 @@ on:
...
+    name: tox run -e lint
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install tox & uv
+        run: |
+          pipx install tox
+          sudo snap install astral-uv --classic
```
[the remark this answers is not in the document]
marceloneppel: I see no problem if the version from the snap satisfies our needs.

.github/workflows/ci.yaml
marceloneppel: Let's see.

marceloneppel (approved):

.github/workflows/ci.yaml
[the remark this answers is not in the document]
taurus-forever: Let's start with Snap. Bonus: snap is maintained by Canonical.

.github/workflows/ci.yaml
taurus-forever: Discuss it with Carl if we like uv in pgbouncer.
I recall Carl had some concerns about uv and multiarch/multi-branch long-term support.

taurus-forever (approved):

PR #546, opened 2025-05-22
[MISC] Sync lock

dragomirp:
* Manual sync of the uv lock
* Additional focal constraints


pyproject.toml
```diff
@@ -28,7 +28,9 @@ libs = [
     # grafana_agent/v0/cos_agent.py
     "cosl",
     # tls_certificates_interface/v2/tls_certificates.py
-    "cryptography",
+    "cryptography; python_version > '3.8'",
+    # Focal cannot build 45
+    "cryptography < 45; python_version < '3.10'",
```
dragomirp, line 33: Latest cryptography can't be build on focal (due to setuptools-rust and maturin constraints). Should we keep the same version everywhere instead?

marceloneppel: I'd prefer to keep this constraint only here to keep other charms up to date.

marceloneppel (approved):

taurus-forever (approved):
Fair.

P.S. I will bring this to Charming Leads as an another reason to stop supporting Focal OR involve canonical into 10 years setuptools-rust support... :crossed_fingers:

PR #547, opened 2025-05-23
[MISC] TLSv4

dragomirp:


dragomirp:
PGB should stay on tlsv2 for the time being.

PR #554, opened 2025-06-20
[DPE-7609] Load login hook for every connection

marceloneppel:
## Issue
The login hook is not executed when a session is reused from the PgBouncer pool, resulting in the relation user not being escalated to the database owner user in subsequent connections after the first one.

## Solution
Load the login hook extension in every connection, so it functions similarly to the regular login procedure.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -48,8 +48,8 @@
   customManagers: [
     {
       customType: 'regex',
-      fileMatch: [
-        '^\\.github/workflows/[^/]+\\.ya?ml$',
+      managerFilePatterns: [
+        '/^\\.github/workflows/[^/]+\\.ya?ml$/',
       ],
       matchStrings: [
         '(libjuju: )==(?<currentValue>.*?) +# renovate: latest libjuju 2',


taurus-forever (approved):

tests/integration/relations/pgbouncer_provider/test_pgbouncer_provider.py
```diff
@@ -151,6 +151,10 @@ async def test_database_version(ops_test: OpsTest):
 
 async def test_database_admin_permissions(ops_test: OpsTest):
     """Test admin permissions."""
+    if os.environ["POSTGRESQL_CHARM_CHANNEL"].split("/")[0] != "14":
+        pytest.skip(
+            "Skipping check for database and user creation permissions on PostgreSQL above 14, as they are not supported."
+        )
```
marceloneppel, line 157: The creation of a database in this test was passing before because the relation user has the `CREATEROLE` database role (https://github.com/canonical/postgresql-test-app/blob/82e88820bd35dd368a15c1ffcd3dcb91656fac48/src/charm.py#L33).

It was missed to block this permission on the PostgreSQL library (https://github.com/canonical/pgbouncer-operator/blob/74a988e3707a068e7bbc5645cce2e36402995a7a/lib/charms/postgresql_k8s/v1/postgresql.py#L774-L775). We'll need to update the test app first to not request that (and also to request the `charmed_database_owner` extra user role instead of the `CREATEDB` attribute) and them fix the library to not allow those attributes to be requested.

taurus-forever:
@7annaba3l Mohamed, as discussed on Friday, we need your approve to merge/release this fix.

PR #570, opened 2025-09-01
[DPE-7993] Admin console access

dragomirp:
Create a user for admin console access


src/relations/backend_database.py
```diff
@@ -159,16 +161,28 @@ def collect_databases(self) -> List[str]:
 
         return databases
 
-    def generate_monitoring_hash(self, monitoring_password: str) -> str:
-        """Generate monitoring SCRAM hash against the current backend."""
+    def generate_scram_hash(self, user: str, password: str) -> str:
+        """Generate SCRAM hash against the current backend."""
```
dragomirp, line 165: Since we are hashing multiple users.

taurus-forever (approved):
Nice! Thank you!

PR #577, opened 2025-09-25
[DPE-8435] Update lock and handle init auth exceptions

dragomirp:
* Update uv lock
* Catch possible exceptions when trying to initialise the auth function
* Use `/mnt` for LXC storage


src/relations/backend_database.py
```diff
@@ -237,7 +237,14 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
         ) else self.postgres.create_user(
             self.auth_user, hashed_password, admin=True, database=self.database.database
         )
-        self.initialise_auth_function(self.collect_databases())
+        try:
+            self.initialise_auth_function(self.collect_databases())
+        except Exception as e:
```
dragomirp, line 242: The pgb rel user may not yet have access to the postgres db.

PR #656, opened 2026-04-22
[DPE-9932] TICS with UV and update lock

dragomirp:
* UV deps installation in weekly scan workflow
* Update lock

### DIFF
--- a/.github/workflows/tiobe_scan.yaml
+++ b/.github/workflows/tiobe_scan.yaml
@@ -36,8 +36,9 @@ jobs:
 
       - name: Install additional dependencies for TICS
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/tiobe_scan.yaml
```diff
@@ -36,8 +36,9 @@ jobs:
 
       - name: Install additional dependencies for TICS
         run: |
-          pip install flake8 pylint pytest
```
dragomirp, line 39: Pinned pytest should be installed by UV.

carlcsaposs-canonical (approved):

marceloneppel (approved):

taurus-forever (approved):

PR #661, opened 2026-05-21
[MISC] Limit tls relations to Jammy+

dragomirp:
TLS relation needs the `cryptography` package to work, but for PGB it is only useful when related to `data-integrator` (exposed connection).

Since `data-integrator` is only available for jammy and noble and cryptography version compatible with focal has CVEs, the PR disables TLS relation and doesn't install cryptography when deployed on focal base.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,11 +37,10 @@ libs = [
     # grafana_agent/v0/cos_agent.py
     "cosl",
     # tls_certificates_interface/v2/tls_certificates.py
+    # Jammy+ only
     "cryptography; python_version > '3.9'",
-    # Focal cannot build 45
-    "cryptography <45; python_version < '3.10'",
-    "jsonschema",
-    "idna",
+    "jsonschema; python_version > '3.9'",
+    "idna; python_version > '3.9'",
 ]
 
 format = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,3 +148,11 @@ max-complexity = 10
 
 [tool.ruff.lint.pydocstyle]
 convention = "google"
+
+[tool.ty.environment]
+python = ".tox/lint/"
+extra-paths = ["./lib"]
+
+[tool.ty.src]
+include = ["src"]
+exclude = ["tests"]
--- a/src/charm.py
+++ b/src/charm.py
@@ -902,7 +902,9 @@ class PgBouncerCharm(TypedCharmBase):
             template = Template(file.read())
             databases = self._get_relation_config()
             readonly_dbs = self._get_readonly_dbs(databases)
-            enable_tls = all(self.tls.get_tls_files()) and self._is_exposed
+            enable_tls = (
+                sys.version_info > (3, 9) and self._is_exposed and all(self.tls.get_tls_files())
+            )
             addr = "*" if self._is_exposed else "127.0.0.1"
             # Modify & render config files for each service instance
             for service_id in self.service_ids:
--- a/src/charm.py
+++ b/src/charm.py
@@ -13,6 +13,7 @@ import platform
 import pwd
 import shutil
 import subprocess
+import sys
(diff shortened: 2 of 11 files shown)


src/charm.py
```diff
@@ -34,7 +35,9 @@
 from charms.operator_libs_linux.v2 import snap
 from charms.pgbouncer_k8s.v0.pgb import generate_password
 from charms.postgresql_k8s.v0.postgresql import PERMISSIONS_GROUP_ADMIN
-from charms.postgresql_k8s.v0.postgresql_tls import PostgreSQLTLS
+
+if sys.version_info > (3, 9):
+    from charms.postgresql_k8s.v0.postgresql_tls import PostgreSQLTLS
```
dragomirp, line 40: Conditional import.

src/charm.py
```diff
@@ -186,11 +195,18 @@ def __init__(self, *args):
             substrate="vm",
         )
 
-    @property
+    def _on_tls_joined(self, _) -> None:
+        self.unit.status = BlockedStatus("TLS unavailable on 20.04 base")
+
+    def _on_tls_broken(self, _) -> None:
+        if self.is_blocked and self.unit.status.message == "TLS unavailable on 20.04 base":
+            self.unit.status = ActiveStatus()
```
dragomirp, line 203: Block to inform the user.

src/charm.py
```diff
@@ -230,11 +246,6 @@ def create_instance_directories(self):
             os.makedirs(f"{app_run_dir}/{INSTANCE_DIR}{service_id}", 0o777, exist_ok=True)
             os.chown(f"{app_run_dir}/{INSTANCE_DIR}{service_id}", pg_user.pw_uid, pg_user.pw_gid)
 
-    @property
-    def tracing_endpoint(self) -> Optional[str]:
-        """Otlp http endpoint for charm instrumentation."""
-        return self._tracing_endpoint_config
-
```
dragomirp, line 237: Doesn't seem to be used anywhere.

src/charm.py
```diff
@@ -155,7 +158,13 @@ def __init__(self, *args):
         self.client_relation = PgBouncerProvider(self)
         self.legacy_db_relation = DbProvides(self, admin=False)
         self.legacy_db_admin_relation = DbProvides(self, admin=True)
-        self.tls = PostgreSQLTLS(self, PEER_RELATION_NAME)
+        if sys.version_info > (3, 9):
+            self.tls = PostgreSQLTLS(self, PEER_RELATION_NAME)
+        else:
+            self.tls = None
```
dragomirp, line 164: For easier checking if TLS is available or not.

pyproject.toml
```diff
@@ -149,3 +148,11 @@ max-complexity = 10
...
 [tool.ruff.lint.pydocstyle]
 convention = "google"
+
+[tool.ty.environment]
+python = ".tox/lint/"
+extra-paths = ["./lib"]
+
+[tool.ty.src]
+include = ["src"]
+exclude = ["tests"]
```
dragomirp, line 158: Scaffolding for ty. Not yet enabled.

dragomirp:
We'll most probably need a spec for this.

src/charm.py
```diff
@@ -186,11 +195,18 @@ def __init__(self, *args):
             substrate="vm",
         )
 
-    @property
+    def _on_tls_joined(self, _) -> None:
+        self.unit.status = BlockedStatus("TLS unavailable on 20.04 base")
```
marceloneppel, line 199: I think we should check for this message in other relation events too, like the backend database joined event, where this message will be replaced by the Active Status.

marceloneppel (changes requested):

taurus-forever (approved):
LGTM.

The Focal TLS restrictions must be represented in the next pgbouncer release notes.

PR #663, opened 2026-05-27
[MISC] Update deps, conditionally check for otel distros

dragomirp:
* Update Python dependencies
* Drop locks for Python 3.9
* Only run otel cleanup for focal


src/charm.py
```diff
@@ -13,18 +13,19 @@
 import pwd
 import shutil
 import subprocess
+import sys
 from configparser import ConfigParser
 from typing import Dict, List, Literal, Optional, Union, get_args
 
-from utils import _remove_stale_otel_sdk_packages
+if sys.version_info < (3, 9):
+    from utils import _remove_stale_otel_sdk_packages
```
dragomirp, line 21: Focal is stuck to an older otel version and still needs the cleanup.

src/utils.py
```diff
@@ -6,9 +6,13 @@
...
 import os
 import shutil
+import sys
 from collections import defaultdict
 
-from importlib_metadata import distributions
+if sys.version_info < (3, 9):
+    from importlib_metadata import distributions
+else:
+    distributions = None
```
dragomirp, line 15: For test patching.

pyproject.toml
```diff
@@ -12,18 +12,18 @@ charm = [
     "jinja2==3.1.6",
     "ops[tracing]==3.7.0; python_version > '3.9'",
     # Focal cannot use 3
-    "ops[tracing]<3; python_version < '3.10'",
+    "ops[tracing]<3; python_version == '3.8'",
```
dragomirp, line 15: To make uv not resolve the dependency for 3.9. We don't have a charm using 3.9 and packages are dropping support, so try to maintain the lock only for the Python EOL version we need.

carlcsaposs-canonical (approved):

marceloneppel (approved):

PR #669, opened 2026-06-08
[MISC] Update deps backoff

dragomirp:
Follow up to https://github.com/canonical/data-platform/pull/42.

* Add minimum age for uv lock
* Enable pep621 manager for renovate

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -3,6 +3,11 @@
   extends: [
     'github>canonical/data-platform//renovate_presets/charm.json5',
   ],
+  enabledManagers: [
+    'pep621',
+    'github-actions',
+    'custom.regex',
+  ],
   reviewers: [
     'team:data-postgresql',
   ],
--- a/tox.ini
+++ b/tox.ini
@@ -36,7 +36,8 @@ description = Check code against coding style standards
 commands_pre =
     uv --config-file=tox_uv.toml sync --active --group lint --group format
 commands =
-    uv lock --check
+    # Lock checking doesn't work with exclude newer: https://github.com/astral-sh/uv/issues/18775
+    # uv lock --check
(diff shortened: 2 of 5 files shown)


taurus-forever (dismissed):

.github/renovate.json5
```diff
@@ -3,6 +3,11 @@
   extends: [
     'github>canonical/data-platform//renovate_presets/charm.json5',
   ],
+  enabledManagers: [
+    'pep621',
```
dragomirp, line 7: To delegate at least the lock update to renovate.

tox.ini
```diff
@@ -36,7 +36,8 @@ description = Check code against coding style standards
 commands_pre =
     uv --config-file=tox_uv.toml sync --active --group lint --group format
 commands =
-    uv lock --check
+    # Lock checking doesn't work with exclude newer: https://github.com/astral-sh/uv/issues/18775
+    # uv lock --check
```
carlcsaposs-canonical: reading the issue, I'm a bit confused why this affects us

not sure if I'm understanding the issue—it looks like it happens when uv is run in two different places with different config (i.e. one of the places does not have exclude-newer set in pyproject.toml, for example)

do you know why we're affected by the issue?

marceloneppel (approved):

tox.ini
dragomirp: Looks like the cause was passing `tox_uv.toml` to `uv sync` which removes the `exclude-newer` directives from the lock.

tox_uv.toml
carlcsaposs-canonical, line 1: maybe instead of tox_uv.toml, we can remove `no-binary` from pyproject.toml and set UV_NO_BINARY in charmcraft.yaml? so that we only need to maintain 1 uv config

carlcsaposs-canonical (approved):

PR #685, opened 2026-07-07
Re-enable unstable test to gather the current test status

taurus-forever:
## Issue

One test was disable for a while

## Solution

Let's re-enable it to collect the current stats for the test
(after the recent and  major stabilization for PG and PGB).

### DIFF
--- a/tests/integration/relations/pgbouncer_provider/test_pgbouncer_provider.py
+++ b/tests/integration/relations/pgbouncer_provider/test_pgbouncer_provider.py
@@ -377,7 +377,6 @@ async def test_scaling(ops_test: OpsTest):
         )
 
 
-@pytest.mark.skip(reason="Unstable")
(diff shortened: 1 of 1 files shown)


taurus-forever:
@dragomirp it looks like this is still unstable:
https://github.com/canonical/pgbouncer-operator/actions/runs/28855418614/job/85627326354?pr=685#step:6:2602

```
  File "/root/spread_project/tests/integration/helpers/postgresql_helpers.py", line 103, in check_database_users_existence
    assert user not in output
           ^^^^^^^^^^^^^^^^^^
AssertionError
```

Not-long-enough waiting?

taurus-forever:
Reproduced the issue locally, bugreported as https://github.com/canonical/pgbouncer-operator/issues/687 
Reproduced 3/3 times (100%).

We will consider to re-enable the test once the issue is addressed. .

Issue #690, opened 2026-07-10
ERROR: DISCARD ALL cannot run inside a transaction block due to hardcoded server_reset_query_always=1


jansdhillon:
## Problem

When pgbouncer is configured in transaction mode (`pool_mode = transaction`), the charm hardcodes `server_reset_query_always = 1`, which causes pgbouncer to run `DISCARD ALL; LOAD 'login_hook';` on every transaction recycle. This results in continuous errors in the PostgreSQL logs:

```
ERROR: 25001: DISCARD ALL cannot run inside a transaction block
```

pgbouncer recovers from these errors gracefully and continues serving queries. Transaction mode is functional but produces log noise and connection churn on every transaction boundary, which may degrade performance under load.

## Impact

Operators are pushed toward session mode to avoid the errors. In session mode, pgbouncer is a pure proxy with no connection pooling benefit. We tested transaction mode on our staging environment and observed:

- Errors fire on every connection recycle but pgbouncer recovers automatically
- Connection count dropped from ~233 (session mode) to ~165 (transaction mode) under the same workload
- No user-visible failures observed in initial testing, but full load testing has not been completed

## Fix

Expose `server_reset_query_always` as a configurable option (defaulting to `0`) so operators can disable the `LOAD 'login_hook';` reset query when not needed by their application.

## Workaround

Manually edit the pgbouncer config on each unit to change `server_reset_query = DISCARD ALL; LOAD 'login_hook';` to `server_reset_query = DISCARD ALL;` and remove `server_reset_query_always = 1`. This gets overwritten on charm hooks.

## Related

- Same issue filed against K8s operator: canonical/pgbouncer-k8s-operator#740 / [DPE-9936](https://warthogs.atlassian.net/browse/DPE-9936)
- Observed on Landscape SaaS PS7 deployment (canonical/landscape-saas)

[DPE-9936]: https://warthogs.atlassian.net/browse/DPE-9936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

cengizcan-canonical:
Same errors here with `pool_mode = session`. So not limited to transaction mode.

```
pool_mode = session
server_reset_query = DISCARD ALL; LOAD 'login_hook';
server_reset_query_always = 1
```

```
[...]

PR #711, opened 2026-07-29
[DPE-10646] Conditionally add discard directives

dragomirp:
Relates to https://github.com/canonical/pgbouncer-operator/issues/690

## Issue
When adapting the charm to Postgresql 16 charm roles management, we added a workaround for recycled connections, so that the `charmed_{dbname}_owner` role is not lost. The workaround generates excessive error logs in the Postgresql charm and does not properly recycle the relation.

## Solution
The workaround is not necessary for Postgresql 14 charm, since roles work differently and for Postgresql 16 charm it should match the discard command without resetting the role.

### DIFF
--- a/templates/pgb_config.j2
+++ b/templates/pgb_config.j2
@@ -42,6 +42,5 @@ client_tls_cert_file = {{ cert_file }}
 client_tls_sslmode = prefer
 {% endif %}
 {% if backend_version > 16 -%}
-server_reset_query = DISCARD ALL; LOAD 'login_hook';
-server_reset_query_always = 1
+server_reset_query = CLOSE ALL; RESET ALL; DEALLOCATE ALL; UNLISTEN *; SELECT pg_advisory_unlock_all(); DISCARD PLANS; DISCARD TEMP; DISCARD SEQUENCES;
 {% endif %}
--- a/templates/pgb_config.j2
+++ b/templates/pgb_config.j2
@@ -41,5 +41,7 @@ client_tls_ca_file = {{ ca_file }}
 client_tls_cert_file = {{ cert_file }}
 client_tls_sslmode = prefer
 {% endif %}
+{% if backend_version > 16 -%}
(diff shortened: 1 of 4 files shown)


templates/pgb_config.j2
```diff
@@ -41,5 +41,6 @@ client_tls_ca_file = {{ ca_file }}
 client_tls_cert_file = {{ cert_file }}
 client_tls_sslmode = prefer
 {% endif %}
-server_reset_query = DISCARD ALL; LOAD 'login_hook';
-server_reset_query_always = 1
+{% if backend_version > 16 -%}
```
dragomirp, line 44: Assuming that the behaviour would be similar for PG 18. We may need to revisit since `login_hook` is deprecated in favour of login event trigger for PG 17+.

templates/pgb_config.j2
```diff
@@ -41,5 +41,6 @@ client_tls_ca_file = {{ ca_file }}
 client_tls_cert_file = {{ cert_file }}
 client_tls_sslmode = prefer
 {% endif %}
-server_reset_query = DISCARD ALL; LOAD 'login_hook';
```
dragomirp, line 44: `DISCARD ALL` should be equivalent to:
```
CLOSE ALL;
SET SESSION AUTHORIZATION DEFAULT;
RESET ALL;
DEALLOCATE ALL;
UNLISTEN *;
SELECT pg_advisory_unlock_all();
DISCARD PLANS;
DISCARD TEMP;
DISCARD SEQUENCES;
```

So running all of those explicitly except `SET SESSION AUTHORIZATION DEFAULT`.

Postgresql documentation: https://www.postgresql.org/docs/16/sql-discard.html

templates/pgb_config.j2
taurus-forever: Created https://warthogs.atlassian.net/browse/DPE-10769 for PG18.

taurus-forever (approved):

marceloneppel (approved):

